PR #7, opened 2025-12-10
feat: Adjust patching progress animation

MarcaDian:


MarcaDian:
The last stage (Writing the APK) for YouTube takes a lot of time, but for Music it's fast, so I don't know how to make them reach 99% differently. I tried adding adaptive progress, but this significantly complicates the code without benefit.

LisoUseInAIKyrios:
I put in a temporary workaround where steps can have sub steps. The sub steps basically increase the weight of that step so resource compilation/signing count more than a single patch does.

But, it could be refactored more so steps can be a percentage. That would allow the final steps to always count for 30-40% of the total time, and patching would be 40%, with decoding being 20-30%.

LisoUseInAIKyrios:
This looks a lot smoother and has way better feedback.

But I kind of liked the excitement of seeing the percentage numbers pause for small moments then it blasts up in steps.

Could the progress bar have two animation modes, a slow crawl (always present), and when a step does complete the circle animation quickly catches up to where the progress actually is?  So it still has the excitement of progress spurts, but the circle progress bar is never frozen.

LisoUseInAIKyrios:
Looks good. Feel free to adjust the weights more.

YT still shows a "this step may take a while" on my 1 year old phone that is somewhat fast, but it does explain why the last 10 percent suddenly got a lot slower. I think changing the steps to a percentage weighting might help. Could do that later (or never).

LisoUseInAIKyrios:
I think this can be simplified using a timer that is called every second, and the animation is then handled using the existing compose code. Will push a change once I have the smaller details figured out.

MarcaDian:
I've prepared changes to break our files into simpler components, based on this branch without your latest commit.

LisoUseInAIKyrios:
Ok I'm done with changes here.

You may want to adjust the constants in the `overEstimateProgressAdjustment()` if it's hitting 98% too soon on some devices.

Otherwise this seems good. It's not overly complicated and it fixes the usability issue of showing a frozen progress bar.

MarcaDian:
Can be closed as changes has moved to https://github.com/MorpheApp/morphe-manager/pull/8

PR #8, opened 2025-12-10
feat: Add adaptive landscape mode

MarcaDian:
Changes:
- The interface adapts when rotate the screen 
- Bulky files are divided into simpler components for easier development
- Adjust patching progress animation

### DIFF
--- a/app/src/main/java/app/revanced/manager/ui/component/morphe/settings/AppearanceSection.kt
+++ b/app/src/main/java/app/revanced/manager/ui/component/morphe/settings/AppearanceSection.kt
@@ -23,6 +23,7 @@ import app.morphe.manager.R
 import app.revanced.manager.ui.theme.Theme
 import app.revanced.manager.ui.viewmodel.GeneralSettingsViewModel
 import app.revanced.manager.util.toColorOrNull
+import kotlinx.coroutines.CoroutineScope
 import kotlinx.coroutines.launch
 
 /**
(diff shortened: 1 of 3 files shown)


LisoUseInAIKyrios:
This is your PR. Feel free to squash merge to dev when you think these are done for now.

I will make a separate PR to finish root mode off this after this is merged.

MarcaDian:
I want to make some changes to the theme settings. I will write when it is ready, because I am afraid to press such buttons))

LisoUseInAIKyrios:
This branch has a bunch of refactoring and it would be best to merge so my changes don't collide.

Squashing is easy (and we can fix if something is goofed).

Click down arrow beside "Merge pull request", change to squash and merge, click "confirm squash".

LisoUseInAIKyrios:
I'm  not sure if it was refactoring, or URV 1.6.0 changes, but on a new installation when either app is tapped it shows "No patch can be found. Check your patches". The patches have been updated but the UI seems to be using the initial app state.

It only happens on first launch of a clean install or after clearing the app data.

I think refactoring some code into functions broke the `by` syntax and it's not passing a pointer object but instead only passing the current value of the fetch and the number of patches that exist.

LisoUseInAIKyrios:
home page

MarcaDian:
It seems to me that this feature (auto-update on startup) does not work correctly in URM, during the first launch, it shows a dialog in which need to select auto-update, I have seen similar behavior many times when trying to fix the showing and fetching of updates.

Maybe we should not use the built-in auto-update code.

For example, try using the code in the Update button in the bundle card or the pre-release switch in the settings.

LisoUseInAIKyrios:
It looks like it is fetching. The screenshot shows the correct patches release but it doesn't recognize the number of patches.

I agree that making an app startup method call to fetch bundles makes the most sense.

MarcaDian:
I tried to test the latest changes, maybe it's because I changed the orientation during patching (the progress bar is redrawn at this time)

<details>

## Then




## Then




</details>

Edit: No, this is not related to changing the screen orientation.

MarcaDian:
Also, the function of getting patches after switching pre-releases (or Check button) is broke, I think this is due to changes to the `private inner class Update()` from the `PatchBundleRepository class`.

PR #10, opened 2025-12-11
feat: Morphe homepage root installation

LisoUseInAIKyrios:
Finished prior work of adding root install in the simple homepage.

### DIFF
--- a/app/src/main/java/app/revanced/manager/ui/component/morphe/home/MorpheDialogs.kt
+++ b/app/src/main/java/app/revanced/manager/ui/component/morphe/home/MorpheDialogs.kt
@@ -5,7 +5,6 @@ import android.content.res.Configuration
 import androidx.compose.foundation.background
 import androidx.compose.foundation.layout.*
 import androidx.compose.foundation.rememberScrollState
-import androidx.compose.foundation.shape.CircleShape
 import androidx.compose.foundation.shape.RoundedCornerShape
 import androidx.compose.foundation.verticalScroll
 import androidx.compose.material.icons.Icons
@@ -85,6 +84,7 @@ private fun Modifier.dialogHeight(): Modifier {
  * Supports scrollable content with fixed header and footer
  *
  * @param onDismissRequest Called when user dismisses the dialog
+ * @param title Optional title - stays fixed at top
  * @param header Optional header content (icon + title) - stays fixed at top
  * @param footer Optional footer content (buttons) - stays fixed at bottom
  * @param content Scrollable dialog content
@@ -92,6 +92,7 @@ private fun Modifier.dialogHeight(): Modifier {
 @Composable
 fun MorpheDialog(
     onDismissRequest: () -> Unit,
+    title: String? = null,
     header: (@Composable () -> Unit)? = null,
     footer: (@Composable () -> Unit)? = null,
     content: @Composable () -> Unit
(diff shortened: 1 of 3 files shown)


LisoUseInAIKyrios:
The only way to know a device is rooted, is to do something that requires root, and magisk/etc prompts the user to grant root.

So, that makes it tricky presenting a "Use root installation" UI switch in the simple settings when root status has not been confirmed.

The choices are:
1. Always show a "Use root installation" switch in the simple settings, with the downside that the 99% of users who don't have root will be presented with a setting that cannot be used.
2. Don't show any simple mode root settings until root is granted. But try to request root, either when patching starts or when they open the simple settings menu. If the user grants root access, then show the "use root installation" switch in the simple settings, and unless the user turns off the switch then also install with root by default.
3. Don't show any root options in the simple settings, and don't automatically request root access. To use root the user must go into the advanced settings menu and change the primary installer to root, then approve the magisk prompt.

I don't like option 1 because it shows settings that are meaningless and potentially confusing to 99% of users.

Option 2 is by far the easiest and most seamless for the end user. But is it rude requesting root preemptively? The user can deny the magisk popup, and it would only ask once ever. But it is concerning the app is requesting root without the user knowing why.

Option 3 is the safest/laziest. We don't request root unless the user opts in using the advanced settings and the user changes the installer type to root.


I am in favor of option 2 or option 3. Option 3 is the lowest code maintenance and less likely to have future bugs.

LisoUseInAIKyrios:
I'm going to do option 3 because it's the simplest, root users are rare, and non-root install can change the app icon.

Simple install mode will still show the slightly different instructions that were previously added.

LisoUseInAIKyrios:
@MarcaDian If you want to make any adjustments to the mount (root) dialogs or anything else about root usage, then feel free. Otherwise I think this is ready to merge.

MarcaDian:
I'm some busy at real job, so I'll check later tonight.

Has the second dialog been fixed? In the previous code it showed the same instructions for root users as for non-root users. Can you write here the steps of the instructions?

Also, might need to use a different solution for landscape mode instead of scrolling.
We could try scrolling the content in the dialog instead of the entire dialog, or try adapting the text/content size based on the height.

The hard way is to completely disable landscape mode and remove the adaptation code, but this might be a problem for those who will use the manager on non-standard devices such as TVs or tablets.

LisoUseInAIKyrios:
All the dialogs show the correct information for root and non root. I'll post screenshots later.

Can show root UI by changing `usingMountInstall` to true:
https://github.com/MorpheApp/morphe-manager/blob/65ba674881b0c2b7689dba475b32a4ece43b8ef2/app/src/main/java/app/revanced/manager/ui/screen/MorpheHomeScreen.kt#L68-L69

MarcaDian:
?

Maybe the buttons should be made in 1 row in landscape view

https://github.com/user-attachments/assets/711023f8-7799-42b5-9480-45863523fdda

LisoUseInAIKyrios:
maybe remove the cancel button entirely? The user can press back button if they must.

LisoUseInAIKyrios:
Maybe remove the icon from the top too. Keep everything on screen as minimal as possible so there's less to read/process.

LisoUseInAIKyrios:
From what I can find, it might be possible for Manager to monitor the device downloads and automatically check for the expected APK. It only works with more recent Android releases, but it seems that if a criteria is given (such as apk files), then files in the download folder can be scanned without requiring approval from the end user.

This could remove the need for the user to pick the file when they return to the app, because Manager would detect the correct apk (correct version number, and could check the download date to verify it's new and not some old patched apk).,

Maybe it was A.I. hallucinating this feature, but could explore this more at a later time.

MarcaDian:
I think it's ready (I didn't check the root layout)

Also I didn't notice at what point the patch update function broke.

LisoUseInAIKyrios:
I have patch update almost fixed in the other PR. It broke with 1.6.0 update.

PR #11, opened 2025-12-11
fix: Resolve bundle fetching from upstream merge

MarcaDian:


LisoUseInAIKyrios:
It seems to be working now. It was failing before because it was trying to fetch from the non-existent API source. I commented out everything to do with that

LisoUseInAIKyrios:
I just noticed now, but on a clean install if an app button is pressed it says "No patches found" when it previously said something like "Updating patches"

Edit: fixed

LisoUseInAIKyrios:
Pre-release isn't working. It's still uses dev.4 and not dev.5

Probably something simple

PR #13, opened 2025-12-13
feat: Adjust layout of patches list and change log in modal patches bundle

MarcaDian:


LisoUseInAIKyrios:
It's an issue with upstream Manager, but the change logs in Manager Compose only shows the changes for that exact patch release. So users who only infrequently repatch have no idea everything that changed. Manager Flutter automatically showed all previous change logs since the last time patched, so Compose is a step down in this regard.

When clicking on the patches version, I think we could show the entire changelog.md file, such as https://github.com/MorpheApp/morphe-patches/blob/dev/CHANGELOG.md 

Then users can scroll thru all the changes. If we really wanted to get fancy we could detect what version was last patched (URV stores that information) just like Manager Flutter, but the simple change is just show the entire file so we don't have more deeply integrate with URV.

We would need to add a field to the patches bundle json to specify the changelog url, but that would be easy.

We can do this at a later time, because we have enough to finish right now.

MarcaDian:
The only thing I can't fix is ​​that the bottomsheet handle bar is displayed below the status bar (edge-​​to-edge mode). I've tried adding different insets, updating the compose and material versions, but to no avail.

PR #16, opened 2025-12-15
chore: Sync strings from Crowdin

AAMATT:


LisoUseInAIKyrios:
There is no way to sync from Crowdin to GitHub unless it opens/updates a PR. That's going to be a lot of spam if we sync often. We must sync often to keep the history of changes from Crowdin since we're using a free account.

I will setup a GitHub separate sync action to quietly pull the latest to GitHub, and then to merge translations we can use an action to squash the crowdin branch to dev.

LisoUseInAIKyrios:
I'm going to use regular old GitHub actions. This Crowdin GitHub plugin is lacking.

LisoUseInAIKyrios:
I kept the existing Ukraine translations and I'll replace them when this is fully setup

PR #17, opened 2025-12-15
feat: UI & UX Improvements

MarcaDian:
### Changes

1. Display Install / Mount / Uninstall buttons directly on the patcher screen instead of using dialogs
2. Refine the Home screen layout and redesign main buttons
3. Add haptic feedback to main buttons
4. Add animated background to additional screens
5. Add modern fullscreen semi-transparent dialogs instead of legacy dialogs
6. Add Morphing Shape FAB style
7. Introduce more animated background styles:
      - Circles (default)
      - Rings
      - Waves
      - Particles
      - Shapes


LisoUseInAIKyrios:
I only now noticed this PR.

This looks great.

MarcaDian:
I still don't like this solution (I think the upstream installation system is wrong), and we rely on their solution to try to implement ours. 

I'm also trying to clear the local gradle cache, maybe this will solve the delay in opening the settings, the dev2 build from the release doesn't have this problem. And if I build the same version locally, the delay appears.

MarcaDian:
Currently, this installing system does not correctly understand package conflicts. If there is a conflict, the system waits for a timeout of 15 seconds and issues a general error, as before, I had to make a crutch to show this message as a package conflict.

But somehow it works 🤷

LisoUseInAIKyrios:
There has been some upstream changes related to installers. I'll try making another branch and incorporate the changes.

MarcaDian:
A little later I will upload changes that add other types of animated backgrounds for all screens.

MarcaDian:
Gravity animation

<details>




</details>

LisoUseInAIKyrios:
I like the animation options. But maybe the Morphe logos should not be an option? Showing the logo rotated or upside down affects user brand perception of user recollection of the brand.

Maybe it's my device, but the accent colors on the picker are very dark and visually they all look very similar. I can see that is the background color, but maybe for the settings buttons it should show a brighter shade of color to make it easier to compare?

<details>
accent color
</details>

MarcaDian:
Okay, I'll remove the logo option, although I like it the most.
The user opens the application, but apart from the splash animation, the user doesn't see any branding in the UI, except for the About dialog, I thought of changing this perception somehow.
Therefore, I also added a gradient in the main buttons that transitions from the official application colors to the morphe colors. It's like morphing) Maybe it's just in my perception, and no one will notice the hidden meaning.

MarcaDian:
> Currently, this installing system does not correctly understand package conflicts. If there is a conflict, the system waits for a timeout of 15 seconds and issues a general error, as before, I had to make a crutch to show this message as a package conflict.

@LisoUseInAIKyrios I guess I solved this problem by using our solution. For non-root it works as it should, can you check the root installation?

LisoUseInAIKyrios:
I just noticed Manager is spamming the logs hundreds of times per second. This may be causing slowdowns. The background animation logic may need an adjustments.

Edit: This only happens on the home screen, and not while patching.

```
View                    app.morphe.manager                   I  setRequestedFrameRate frameRate=NaN, this=androidx.compose.ui.platform.AndroidComposeView{7867ddb VFED..... ........ 0,0-1080,2340 aid=1073741824}, caller=android.view.ViewGroup.setRequestedFrameRate:9952 androidx.compose.ui.platform.Api35Impl.setRequestedFrameRate:3822 androidx.compose.ui.platform.AndroidComposeView.dispatchDraw:2114 android.view.View.draw:26345 android.view.View.updateDisplayListIfDirty:25175 
View                    app.morphe.manager                   I  setRequestedFrameRate frameRate=NaN, this=android.view.View{a0b1159 V.ED..... ........ 0,0-0,0}, caller=androidx.compose.ui.platform.Api35Impl.setRequestedFrameRate:3822 androidx.compose.ui.platform.AndroidComposeView.dispatchDraw:2115 android.view.View.draw:26345 android.view.View.updateDisplayListIfDirty:25175 android.view.ViewGroup.recreateChildDisplayList:4794 
View                    app.morphe.manager                   I  setRequestedFrameRate frameRate=NaN, this=androidx.compose.ui.platform.AndroidComposeView{7867ddb VFED..... ........ 0,0-1080,2340 aid=1073741824}, caller=android.view.ViewGroup.setRequestedFrameRate:9952 androidx.compose.ui.platform.Api35Impl.setRequestedFrameRate:3822 androidx.compose.ui.platform.AndroidComposeView.dispatchDraw:2114 android.view.View.draw:26345 android.view.View.updateDisplayListIfDirty:25175 
```

LisoUseInAIKyrios:
Root install works.

MarcaDian:
> I just noticed Manager is spamming the logs hundreds of times per second. This may be causing slowdowns. The background animation logic may need an adjustments.

AI writes "Just ignore these logs in development - they don't affect performance and disappear in production builds."

LisoUseInAIKyrios:
Ok it seems to be something specific to the Android Emulator (and is reported to be [harmless even though annoying](https://issuetracker.google.com/issues/448916155)). I'm not seeing the spam log on my device.

MarcaDian:
Do you have the dev version on your phone? Because for example, there is a delay when opening the settings on my phone if the debug version is installed through Android Studio, but if the release version is installed, everything works smoothly and quickly.

LisoUseInAIKyrios:
Android Studio installs debug build. After you mentioned it, I do notice the settings menu animation has a slight delay when sliding out the UI. It might be the spam logger issue, or some other debug environment checks Compose is doing.

But if I build release locally (`./gradlew assembleRelease` then `adb install app/build/outputs/apk/release/*.apk`), then there is no UI lag and it works fine.

MarcaDian:
<details>







</details>

LisoUseInAIKyrios:
The "no help me find it" can be the bottom/right option

LisoUseInAIKyrios:
How does Manager update itself? I assume it's an API end point? Is that present in URV or do we need to figure out something?

For now we can use a url redirect (same as patches, and soon to be for APK web searches)

LisoUseInAIKyrios:
For the background animations, it would be cool to include a "chemistry" option and it shows [Belousov-Zhabotinsky reactions](https://youtu.be/4rYX4ROqRlg) with Morphe colors. It's also slightly fitting because the reaction became more famous thanks to YouTube.

MarcaDian:
I planned to show an manager update dialog and an additional floating button next to the bundle button at startup, but that was at the very beginning, I'm not even sure if it still works.

MarcaDian:
Yes, URV has an autoupdate manager feature, although I've never seen it. Should try installing the old version and see if a new one is available.

MarcaDian:
Yes

<details>




</details>

LisoUseInAIKyrios:
I think it's using GitHub API. Which may have usage limits. I suppose later we can find out if it works, and change to something else if needed.

MarcaDian:
It seems to me that when changing the `usePreRelease` toggle, the bundle is not updated in the background.

Changing the Bundle UI now happens immediately without having to close or open the bottomsheet, by clicking the Check button

LisoUseInAIKyrios:
There is still a minor issue where if pre-release is enabled then the bundle menu is opened, the refresh button looks normal when it should show the spinning progress spinner until the update finishes. Very minor.

MarcaDian:
Does the automatic update of patches count and version in the bundle menu work for you after clicking the prerelease switch in the settings? It doesn't work for me.

LisoUseInAIKyrios:
If I turn on pre-release in the settings, then immediately open the bundle menu, it does not update the patches version there until the bundle menu is closed then opened again.  It's always been like that for me. I think this is a minor issue.

The update snackbar does correctly show, and patching is blocked until the bundles update (most important part).

MarcaDian:
Are there still any issues left?

MarcaDian:
Maybe in the future we need to reconsider the location of floating buttons on all screens, I don't like that it doesn't have a single style, or at least some concept. Everything is done to own taste and very often the opinion changes))

LisoUseInAIKyrios:
The only must have is to use the API to get the correct APKMirror link. It can be as simple as changing from duckduckgo to our API link. I'll try to do that tomorrow.

I created an issue to try and track what we need to do overall.
https://github.com/MorpheApp/morphe-discussion/issues/6

LisoUseInAIKyrios:
If these changes are done, then can squash this until another PR is made for more changes. It's better to have checkpoints of changes to track what happened.

LisoUseInAIKyrios:
> I like the animation options. But maybe the Morphe logos should not be an option? Showing the logo rotated or upside down affects user brand perception of user recollection of the brand.
> 
> Maybe it's my device, but the accent colors on the picker are very dark and visually they all look very similar. I can see that is the background color, but maybe for the settings buttons it should show a brighter shade of color to make it easier to compare?
> 
> Details

I think my comment got lost in translation. I think the old background colors looked great how they were. My issue was the UI accent picker could use a brighter shade of the color (but still use the darker shade for the background). Because it's easy to see the color difference after picking the color, and my problem was just the small picker square was hard to tell apart from other colors.

Screenshot_20251219_115852_Morphe

MarcaDian:
And what would be the correct solution? These are the colors used as preview and as accent/background

https://github.com/MorpheApp/morphe-manager/blob/91bb4d18e7338a0f313d397eaaa05eafa7df298f/app/src/main/java/app/revanced/manager/ui/component/morphe/settings/AppearanceSection.kt#L324-L356

LisoUseInAIKyrios:
Can the previous darker accent colors be used? The settings buttons could increase the brightness of each color a little (use a function to adjust the brightness), but the background uses the darker color.

MarcaDian:
Are we talking about the `Accent pressets` or the `Theme color`? (according to the screenshot)

MarcaDian:
Look at the latest commit in the dev branch. These colors and code are from upstream. And for me, almost all of these theme colors are applied as different shades of purple or blue.

https://github.com/user-attachments/assets/6cebd18a-2215-4687-98e0-4d5a47f4d4bd

LisoUseInAIKyrios:
> Look at the latest commit in the dev branch. These colors and code are from upstream. And for me, almost all of these theme colors are applied as different shades of purple or blue.
> 
>  Screenrecorder-2025-12-19-16-36-21-335.mp4

I think the darker colors shown looks ok, but it needs more variety, and I think the color buttons could show a brighter shade of the same color so it's easier to tell them apart (the colors on the buttons look almost identical between each other)

LisoUseInAIKyrios:
It might benefit from less color choices. I love how the simple settings are minimal without unnecessary stuff, but the UI theme picker has so many presets and options it's almost overwhelming (an issue with upstream URV). Having just a few presets of the best themes could reduce the settings usability complexity.

MarcaDian:
I tried to make an animated splash screen using only standard animations. Should I add this?

https://github.com/user-attachments/assets/3e27a972-7d71-4a54-bfbc-140f548b6fb9

LisoUseInAIKyrios:
Looks cool. I think the parts of the M (two legs and top stroke) could be smoother around the edges so it looks more blob like when morphing. Can add if you'd like.

LisoUseInAIKyrios:
I've wanted to add a morphing logo to YT custom branding theme but never tried yet. It would use a different animation of the M morphing into a red circle square, and then YT continues on with it's animation.

Or patches could not morph the logo into YT logo (probably better for trademark reasons), and instead it simply fades out the M logo while the YT animation starts behind it. This might be really simple to do.

PR #20, opened 2025-12-18
ci: Move to github changelog fork

inotia00:
Just as [patches-bundle.json](https://github.com/MorpheApp/morphe-patches/blob/dev/patches-bundle.json) is updated in the patches repo, the `app-release.json` in the manager repo is also updated when the release action completes.

Once this PR is merged, we can avoid using the GitHub releases API.

See also: https://github.com/MorpheApp/morphe-patches/pull/15.

### DIFF
--- a/api/.releaserc
+++ b/api/.releaserc
@@ -16,7 +16,13 @@
       }
     ],
     "@semantic-release/release-notes-generator",
-    "@semantic-release/changelog",
+    [
+      "@MorpheApp/changelog", {
+        "releaseJson": {
+          "enabled": false
+        }
+      }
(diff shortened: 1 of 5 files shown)


LisoUseInAIKyrios:
Merge whenever you think they're ready

inotia00:
The PR works as intended.

Release information for the manager can be found at `https://github.com/MorpheApp/morphe-manager/blob/dev/app/app-release.json`.

LisoUseInAIKyrios:
After public release can update manager to use that file to check for updates (and not use GitHub api that it's currently using)

PR #26, opened 2025-12-25
refactor: Use variable percentages for each patching step

LisoUseInAIKyrios:
Changes the progress steps from a fixed number to a percentage.

Now executing the patches always takes 40% and writing the apk always takes 40%


LisoUseInAIKyrios:
The progress over estimation logic could be improved further to over estimate based on the current step (and include a time estimate parameter for each step to better estimate the time it will take). But that will wait until later.

PR #27, opened 2025-12-25
feat: Add in-app patches options

MarcaDian:
# What's New in This PR
## 🎨 Patch Options System

- Added in-app patch options configuration
- Implemented expandable sections for YouTube and YouTube Music
- Fetch patch options and descriptions directly from bundle repository
- Added color picker for theme customization options
- Show waiting message during bundle updates

## 🎄 Seasonal Touch

- Added snow background animation

## 🧹 Code Quality

- Major refactoring following DRY principles
- Removed hardcoded strings
- Simplified and cleaned up codebase
___
<details>
 <summary>How it looks</summary>

https://github.com/user-attachments/assets/668f5683-ed89-402c-8747-6a5a1e33788f
</details>

### DIFF
--- a/app/src/main/java/app/revanced/manager/ui/screen/settings/GeneralSettingsScreen.kt
+++ b/app/src/main/java/app/revanced/manager/ui/screen/settings/GeneralSettingsScreen.kt
@@ -1,5 +1,6 @@
 package app.revanced.manager.ui.screen.settings
 
+import android.graphics.Color.parseColor
 import android.os.Build
 import androidx.annotation.StringRes
 import androidx.compose.foundation.background
@@ -14,7 +15,10 @@ import androidx.compose.material.icons.outlined.Palette
 import androidx.compose.material3.*
 import androidx.compose.runtime.Composable
 import androidx.compose.runtime.getValue
+import androidx.compose.runtime.mutableStateOf
 import androidx.compose.runtime.remember
+import androidx.compose.runtime.saveable.rememberSaveable
+import androidx.compose.runtime.setValue
 import androidx.compose.ui.Alignment
 import androidx.compose.ui.Modifier
 import androidx.compose.ui.draw.alpha
@@ -36,7 +40,9 @@ import app.revanced.manager.ui.component.settings.ExpressiveSettingsCard
 import app.revanced.manager.ui.viewmodel.GeneralSettingsViewModel
 import app.revanced.manager.ui.viewmodel.ThemePreset
 import app.revanced.manager.util.toColorOrNull
+import app.revanced.manager.util.toHexString
 import org.koin.androidx.compose.koinViewModel
+import kotlin.math.roundToInt
 
 val THEME_PRESET_COLORS = listOf(
     Color(0xFF6750A4),
@@ -493,6 +499,8 @@ fun GeneralSettingsScreen(
 }
 
 private data class ThemePresetSwatch(val preset: ThemePreset, @StringRes val labelRes: Int, val colors: List<Color>)
+private data class LanguageOption(val code: String, @StringRes val labelRes: Int)
+
 
 @Composable
 private fun ThemeSwatchChip(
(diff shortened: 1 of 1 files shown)


MarcaDian:
Maybe we should hide some settings in dropdowns so they don't take up too much space, like these. I don't like multi-page interfaces, and I want everything to be on one page, but not too cluttered. That can all be done later.

LisoUseInAIKyrios:
Maybe put the Manager appearance settings into a drop down expandable settings? That would free up a lot of vertical space. If we keep advanced mode for v1.0 then the advanced mode could be it's own button/component.

MarcaDian:
The only problem I'm facing is that we hardcode the options.

For example, for the branding icon selection dialog, we need to show a message about the correct names of these icons. It would be logical to take this directly from the branding patch option. But I'm worried that I won't be able to implement it.

The simplest solution is to add these strings to the manager, but every time we make a change to this patch, we will have to edit them in the manager.

MarcaDian:
Probably this works 🤷 😎

<details>




</details>

LisoUseInAIKyrios:
The patch list contains the patch options. So it should be a matter of parsing it out like RVP/Compose handles it.

LisoUseInAIKyrios:
If we can't get patch options parsing right now, then we can go with what's here with Manager using hard coded patch names. We can later add patch option parsing (and also fix up patcher options to allow indicating an option string is a file/folder path).

LisoUseInAIKyrios:
This is so much easier than Compose or Flutter. An Android file/folder picker is so much easier than being a masochist and manually enter a string for a device folder path.

I think the dialogs should use an opaque background unless a strong blur effect can be used. The text bleeding thru interferes with the readability.

<details>
patch options
</details>

MarcaDian:
Still need to fix the permission to access file system. (In Advanced mode, a request for access is raised)

> I think the dialogs should use an opaque background unless a strong blur effect can be used. The text bleeding thru interferes with the readability.

I haven't done this yet. I'll do it later.

inotia00:
> The only problem I'm facing is that we hardcode the options.
> 
> For example, for the branding icon selection dialog, we need to show a message about the correct names of these icons. It would be logical to take this directly from the branding patch option. But I'm worried that I won't be able to implement it.
> 
> The simplest solution is to add these strings to the manager, but every time we make a change to this patch, we will have to edit them in the manager.

There are two ways. 

One is to parse patches-list.json directly: 

https://github.com/HundEdFeteTree/HappyFunTest/blob/fa6e638a9d3c33bb0f4d624b32fad3d8de160c94/bundles/test-dev-patches-list.json#L1474 

The other is to filter patch options with Morphe Patcher: 

https://github.com/inotia00/revanced-cli/blob/8b432d55364051eb538cfbf81d46666d90befdf3/src/main/kotlin/app/revanced/cli/command/PathesCommand.kt#L48 

https://github.com/inotia00/revanced-cli/blob/8b432d55364051eb538cfbf81d46666d90befdf3/src/main/kotlin/app/revanced/cli/command/OptionsCommand.kt#L52

MarcaDian:
If everything I've implemented works, that'll be cool))

Visually, everything looks correct.

https://github.com/user-attachments/assets/668f5683-ed89-402c-8747-6a5a1e33788f

LisoUseInAIKyrios:
Until Patcher gets support to mark patch option strings as more than just text (a directory, a file path, a hex color, etc), I think we can use the hard coded options logic in this PR.

Because if we parse all the patch options in Manager, we still need to hard code what options are a color, what are a directory, etc. And we'd need to hard code what icons to show for the options.

Patcher options could be modified to include an icon file, or maybe Manager can use the same option icons for the different option data types (color hex string, directory string, file path string, etc).

LisoUseInAIKyrios:
On a new app startup (or after clearing app data), the settings shows an error until the patches are loaded

<details>
new install settings error
</details>

MarcaDian:
This is correct behavior, as currently the patch options are fetched from the patch bundle. It might be more accurate to change this to something like "Wait until the patch bundle is updated"

The func for updating the patch option used in the settings has also been added to the pre-release toggle and the Check button in the sheetbundle

LisoUseInAIKyrios:
URV 1.6.1 has changed the advanced theme options. Should our simple mode reflect that?

It seems there is just a single theme color to pick (called accent color in our picker), and it has a Material You option. I think having just one color picker (just "theme" and not "accent" and "theme") is an improvement.

MarcaDian:
I noticed in URV, the background color selection (Theme color) is not active for me.

<details>




</details>

LisoUseInAIKyrios:
The advanced general color menu also has two different theme previews. The old and the new expressive one. It seems to be left over code because I don't know why it should be shown twice.

LisoUseInAIKyrios:
In advanced mode the "accent" color now seems to only control the foreground components. The background no longer seems to be customizable. I think only setting the foreground color might be ok because it's less complexity. Whatever we do, I think advanced mode color themes should be the same as simple mode (except simple mode will still have it's own specific settings such as background animations).

LisoUseInAIKyrios:
Actually, there is still a "theme color" along with accent color. But it's not selectable as you said.

LisoUseInAIKyrios:
Ok the "theme color" is only selectable when no present is selected. Kind of confusion and not it's clear when no theme is selected.

MarcaDian:
So should we leave it or remove it?

LisoUseInAIKyrios:
What do you think? I think it's simpler without a custom background color (using system dark/light is good enough). That would leave only the "theme presents" (system, light, dark, pure black, material you) and the foreground "accent" color.

We can always later add a custom theme (custom background color) if people really want. But Manager is only used occasionally so having so many superficial visual customization makes the app more complicated without much return.

MarcaDian:
Now let's try how it looks. I don't use it at all)
I will first do and check locally if everything works

MarcaDian:
Ideally, theme presets (background + accent) would be implemented, but I have bad taste in such things, and maybe this is unnecessary.

LisoUseInAIKyrios:
Maybe system/light/dark could be it's own setting and then the foreground/background color is a separate option to pick where we give a few presets (similar to the presets in YT theme option, 'material you' would go here too). Then have no custom color picker at all.

So the settings would be:
- Light dark options:
   - Follow system
   - Light
   - Dark
   - Pure black (uses same dark foreground color of option above, except the background is always pure black)
- Color (changes background and foreground tint. Each preset uses different appropriate color shades for light/dark mode):
   - Material you
   - (whatever preset background/foreground colors presets we pick. Pick as many color pairs we want)


And that's all. No color picker, and only two settings the user can change. I think this could work well.

LisoUseInAIKyrios:
It now works way better 👍

Should Material You be a color preset since it works with light/dark?

I think the "x" preset could be a different icon or changed to only show the color.

MarcaDian:
> I think the "x" preset could be a different icon or changed to only show the color.

X is the lack of an accent color (default), I don't understand what icon can be added

> Should Material You be a color preset since it works with light/dark?

In Advanced mode, Material You is the theme selection item, I did as you suggested at the beginning, do all the same as in Advanced mode.

LisoUseInAIKyrios:
Ok this works then.

MarcaDian:
Not sure if the truncation of the word "Material You" has been resolved, as I can't switch languages, my language uses a long word so it's still hyphenated.
There are two solutions: use another word, such as "Dynamic"?, or disable hyphenation and use "ellipsis" to change word with "Materi..."

<details>




</details>

I don't understand why we even waste time setting up Advanced mode, knowing that the next update will break it all.

LisoUseInAIKyrios:
I hope we don't have to pull from upstream much anymore. It seems all the bugs have been fixed so there may not be any immediate reasons anymore.

We can probably merge our advanced settings changes without too much fuss.

For the text clipping, do whatever is easiest.

I think this PR is about ready?

MarcaDian:
To be honest, I'm already confused by this messy code. It seems that what was planned in this PR at the beginning works, the rest can be fixed later if necessary.

LisoUseInAIKyrios:
Ok. Merge this whenever you want. We can make other changes in another PR.

PR #48, opened 2026-01-03
feat: Add localization to patch options

MarcaDian:
Example

<details>


LisoUseInAIKyrios:
To anyone looking at this and asking "Why are patches option strings in the Manager code?", it's because these strings are temporary until we finish some upcoming patcher changes.

After changes are made, then all temporary strings will be removed from Manager.

LisoUseInAIKyrios:
Can merge whenever ready

MarcaDian:
I'm not sure if this works as intended, but I checked on my locale, and everything seems to be fine.

PR #87, opened 2026-01-08
fix: Allow disabling built-in bundle

LisoUseInAIKyrios:
There has always been redundant delete/disable buttons when navigating into the expert mode bundle, but this change restores the disable button shown in the patch bundle list.

### DIFF
--- a/app/src/main/java/app/revanced/manager/ui/component/bundle/BundleItem.kt
+++ b/app/src/main/java/app/revanced/manager/ui/component/bundle/BundleItem.kt
@@ -386,7 +386,6 @@ fun BundleItem(
                     }
                 }
 
-                // Morphe: For now, don't allow removing the built in patches bundles
                 val toggleIcon = if (src.enabled) Icons.Outlined.Block else Icons.Outlined.CheckCircle
                 val toggleLabel = if (src.enabled) R.string.disable else R.string.enable
                 Box(modifier = Modifier.fillMaxWidth(), contentAlignment = Alignment.Center) {


MarcaDian:
Have you tested this? I'm not sure if there will be any problems since we changed some things.

Next is my opinion:
We left Expert mode only for advanced users. And now almost all open (or any intermediate) problems are related to this mode. I don't understand why users use it, if everything is in Morph mode (except for root options).
Almost every screenshot with the any manager problem I've seen on Reddit (I still can't get used to this social network) is from Expert mode.

LandonMoran:
also adding extra patch bundles is not in normal mode. I believe also being able to change the amount of RAM that the manager can use are settings that are slightly more advanced and might not fit all novice users. But if someone wants them, they might just need one of those. And I'm not sure if switching out of advanced mode actually changes those back or not. In my case, I wanted to use the Shizuku installer, but in order to use it, you have to be in advanced mode, and there are certain settings that are only accessible in advanced mode that aren't that advanced. What I think needs to be done is some of the simpler settings can be moved to the "Not advanced settings" menu and they just have warning labels on them? Or maybe just scrap the advanced mode.

MarcaDian:
If desired, we can transfer the main options, but Liso wants to see as simple settings as possible. In my case, simple settings are enough for me. I'm already tired of this constant "poking" at different settings. That's why I liked the idea of ​​the simplest "patch 1 button" manager

LandonMoran:
> I think either the advanced mode needs to get removed or have more warnings on it. Also, the settings that some users might need that aren't all that advanced should be moved out of advanced mode so that people are not incentivized to turn it on to use them.> Also, adding extra patch bundles is not in normal mode. I believe also being able to change the amount of RAM that the manager can use are settings that are slightly more advanced and might not fit all novice users. But if someone wants them, they might just need one of those. And I'm not sure if switching out of advanced mode actually changes those back or not. In my case, I wanted to use the Shizuku installer, but in order to use it, you have to be in advanced mode, and there are certain settings that are only accessible in advanced mode that aren't that advanced. What I think needs to be done is some of the simpler settings can be moved to the "Not advanced settings" menu and they just have warning labels on them? Or maybe just scrap the advanced mode.

LisoUseInAIKyrios:
I've tested it and it works. The buttons were previously there in URV but commented out. Disabling is needed to patch a PR artifact from the patches repo otherwise the local bundle and the built in bundle both will try to apply and fail.

There still exists disable and delete in the individual package menu on the top bar. I didn't notice these existed and the delete should be removed because if the user deletes the API bundle theres no way to restore it unless they clear the app data.

I agree everyone having issues is using expert mode and right now there is no reason to use it for YT except to enable root install. But since there's already early third party bundles (Reddit) expert mode is already needed less than a week after 1.0.0 release. I'd love for our simple layer to be used for third party bundles and then hide expert mode entirely, but we need to figure out details like multiple patch sources and the UI to pick any app.

MarcaDian:
Just for reference, do you see this? Previously, if there was a "Pending" status, the author did not see it.

image

LisoUseInAIKyrios:
That's a pending review and only the review author sees it until they finish and submit their review.

I made the same mistake when I got started on GitHub ages ago and thought someone was ignoring my comments, lol.

PR #90, opened 2026-01-08
fix: Completely isolate patch options in Morphe and Expert modes

MarcaDian:
Closes https://github.com/MorpheApp/morphe-manager/issues/54


LisoUseInAIKyrios:
Anything left to do here? Ready to merge?

MarcaDian:
I checked that mode isolation works for me, but that user (from pinned issue) is probably misleading me.

LisoUseInAIKyrios:
Then lets merge because it's confirmed locally

PR #93, opened 2026-01-09
feat: Add language picker

MarcaDian:
## Morphe mode

___

### DIFF
--- a/app/src/main/java/app/revanced/manager/ui/component/morphe/home/HomeDialogs.kt
+++ b/app/src/main/java/app/revanced/manager/ui/component/morphe/home/HomeDialogs.kt
@@ -2,6 +2,10 @@ package app.revanced.manager.ui.component.morphe.home
 
 import android.annotation.SuppressLint
 import android.widget.Toast
+import androidx.compose.animation.AnimatedVisibility
+import androidx.compose.animation.core.tween
+import androidx.compose.animation.fadeIn
+import androidx.compose.animation.fadeOut
 import androidx.compose.foundation.layout.*
 import androidx.compose.foundation.shape.RoundedCornerShape
 import androidx.compose.material.icons.Icons
@@ -16,6 +20,7 @@ import androidx.compose.material3.MaterialTheme
 import androidx.compose.material3.Surface
 import androidx.compose.material3.Text
 import androidx.compose.runtime.Composable
+import androidx.compose.runtime.rememberCoroutineScope
 import androidx.compose.ui.Alignment
 import androidx.compose.ui.Modifier
 import androidx.compose.ui.graphics.Color
@@ -36,6 +41,7 @@ import app.revanced.manager.util.htmlAnnotatedString
 import app.revanced.manager.util.toast
 import kotlinx.coroutines.CoroutineScope
 import kotlinx.coroutines.Dispatchers
+import kotlinx.coroutines.delay
 import kotlinx.coroutines.launch
 
 /**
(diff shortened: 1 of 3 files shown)


LisoUseInAIKyrios:
Love the flag emojis. The language picker for patches could do the same.

Should this show a similar message as the patches language setting? 
> Translations for some languages may be missing or incomplete.
>
> To translate new languages or improve the existing translations, visit translate.morphe.software

I just noticed the patches dialog has the shortcut domain url and not https://morphe.software/translate but maybe that's ok because the patches dialog is not a clickable link and the user must manually type out the url. But if we can add an additional dialog button here to manager then it should use https://morphe.software/translate

MarcaDian:
> I just noticed the patches dialog has the shortcut domain url and not

I thought it was intended that way)) 😅

MarcaDian:
By the way, I spent almost a day trying to figure out why my language wasn't changing 🤔 when I selected it from the list, and then I remembered that [you commented on the upstream changes](https://github.com/MorpheApp/morphe-manager/blob/ffa204c4dc4b6c3a0fccc7bb640ff6154decb090/app/src/main/java/app/revanced/manager/MainActivity.kt#L48) 

Jman made these changes specifically for localization purposes.

LisoUseInAIKyrios:
It was intended but then we added https://Morphe.software/translate

So both links exist.

LisoUseInAIKyrios:
The issue with app compat was the app would flash light/dark for a moment on app startup. Might have been only on the first launch on cold start, I don't remember exactly.

LisoUseInAIKyrios:
Some languages are not specific to any country, such as Spanish that's used in dozens of counties. I think showing the flag the language is from is ok (including showing United Kingdom flag for English), but I think the country name isn't needed because the flag already shows that.

MarcaDian:
I think it's because the code for changing the language was commented out and the default language in the code was `English`. I changed it to `System`, and it seems that the light/black screen (which flashes during a fresh install) is gone now.

The artifact will be compiled soon, and anyone who wants to can test it.

MarcaDian:
> but I think the country name isn't needed because the flag already shows that.

This is a system function to get a list of languages. Not sure if this should be changed to a custom list item constructor.

LisoUseInAIKyrios:
I think there's a built in function to get the language name without the country name.

MarcaDian:
Yes, there is a system function to get only the language. I remembered why I did it. Languages like Brazil and Portugal will be called the same and have the same flag, so also  need to show the Country. I'll see if I can combine it.

LisoUseInAIKyrios:
In the hard coded list of languages could use a pair/map to the 2 or 4 character language code (or make a language enum like Patches extension code uses).

So Portuguese and Chinese will have 4 character codes of language and region, and all the others will be only 2 character codes.

MarcaDian:
If anyone wants to test https://github.com/MorpheApp/morphe-manager/actions/runs/20858558454/artifacts/5078361777

app/src/main/res/values/strings.xml
```diff
@@ -131,6 +131,14 @@
     <!-- Appearance Section -->
     <string name="morphe_appearance_options">Appearance options</string>
     <string name="morphe_appearance_options_description">Customize themes, colors, and backgrounds</string>
+    <string name="morphe_appearance_app_language">App Language</string>
+    <string name="morphe_appearance_search">Search</string>
+    <string name="morphe_appearance_no_results">No results found</string>
+    <string name="morphe_appearance_translations_info_title">Translations may be missing</string>
+    <string name="morphe_appearance_translations_info_text">Translations for some languages may be missing or incomplete.
+
+To translate new languages or improve the existing translations, visit</string>
```
LisoUseInAIKyrios:
```suggestion
    <string name="morphe_appearance_translations_info_title">Translations for some languages may be missing or incomplete</string>
    <string name="morphe_appearance_translations_info_text">To translate new languages or improve the existing translations, visit %s</string>
```

It's best to use a parameter for the url, in case a language needs to rearrange the wording.  Wherever this string is used needs to pass in the url as a format parameter

LisoUseInAIKyrios: Oh, I see the url is clickable which may be more difficult to do as a parameter.

MarcaDian:
I haven't come up with anything better than recognizing link text through "."
In this case, it works.

LisoUseInAIKyrios:
that looks great

LisoUseInAIKyrios:
Merge whenever you think it's ready.

MarcaDian:
I'm away from my PC, if everything is fine, you can merge

PR #99, opened 2026-01-10
feat: Add Installer picker to the Morphe settings

MarcaDian:

### DIFF
--- a/app/src/main/java/app/revanced/manager/domain/installer/InstallerManager.kt
+++ b/app/src/main/java/app/revanced/manager/domain/installer/InstallerManager.kt
@@ -47,9 +47,13 @@ class InstallerManager(
         val hiddenPackages = hiddenInstallerPackages
         val entries = mutableListOf<Entry>()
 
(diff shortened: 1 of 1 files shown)


MarcaDian:
Root mode test needed.

LisoUseInAIKyrios:
Root mode works.

Right now it requests root access when the setting menu is first opened, but ideally the "isRooted" check is done only when the installer menu is selected.

LisoUseInAIKyrios:
Something is still requesting root when the simple settings are opened. If this can't be figured out it's ok to use as-is.

LisoUseInAIKyrios:
It's fixed. It now prompts for root only after changing to root in settings menu.

MarcaDian:
I wonder why this wasn't implemented in upstream. I forgot to remove the extra checks for Morphe mode in the Advanced section in Expert, but I'm afraid of breaking something again.
Maybe they'll be around for a while until we reorganize our settings.

LisoUseInAIKyrios:
Merge anytime if this is ready

MarcaDian:
Continued here https://github.com/MorpheApp/morphe-manager/pull/101

PR #107, opened 2026-01-12
feat: Add Expert mode

MarcaDian:
What was done:

- Implemented a new Expert mode interface for advanced users.
- Expanded patch selection options and control in the UI.
- Improved state management and interaction logic for patches.
- Added dialogs for saved apps and patch options.
- Refactored internal code for home and APK management screens.
- Multiple build and cleanup commits to support these changes.
- Added Reddit support.

### DIFF
--- a/app/src/main/java/app/revanced/manager/ui/component/morphe/home/HomeBottomActionBar.kt
+++ b/app/src/main/java/app/revanced/manager/ui/component/morphe/home/HomeBottomActionBar.kt
@@ -1,16 +1,14 @@
 package app.revanced.manager.ui.component.morphe.home
 
+import android.annotation.SuppressLint
 import android.view.HapticFeedbackConstants
 import androidx.compose.foundation.BorderStroke
-import androidx.compose.foundation.layout.Arrangement
-import androidx.compose.foundation.layout.Row
-import androidx.compose.foundation.layout.fillMaxSize
-import androidx.compose.foundation.layout.fillMaxWidth
-import androidx.compose.foundation.layout.height
-import androidx.compose.foundation.layout.padding
-import androidx.compose.foundation.layout.size
+import androidx.compose.foundation.background
+import androidx.compose.foundation.layout.*
+import androidx.compose.foundation.shape.CircleShape
 import androidx.compose.foundation.shape.RoundedCornerShape
 import androidx.compose.material.icons.Icons
+import androidx.compose.material.icons.outlined.Engineering
 import androidx.compose.material.icons.outlined.Settings
 import androidx.compose.material.icons.outlined.Source
 import androidx.compose.material3.CircularProgressIndicator
@@ -20,6 +18,7 @@ import androidx.compose.material3.Surface
 import androidx.compose.runtime.Composable
 import androidx.compose.ui.Alignment
 import androidx.compose.ui.Modifier
+import androidx.compose.ui.draw.clip
 import androidx.compose.ui.graphics.Brush
 import androidx.compose.ui.graphics.Color
 import androidx.compose.ui.graphics.vector.ImageVector
@@ -36,6 +35,8 @@ import app.morphe.manager.R
 fun HomeBottomActionBar(
     onBundlesClick: () -> Unit,
     onSettingsClick: () -> Unit,
+    isExpertModeEnabled: Boolean = false,
+    @SuppressLint("ModifierParameter")
     modifier: Modifier = Modifier
 ) {
     Row(
@@ -54,11 +55,12 @@ fun HomeBottomActionBar(
             modifier = Modifier.weight(1f)
         )
 
-        // Right: Settings button
+        // Right: Settings button with expert mode indicator
         BottomActionButton(
             onClick = onSettingsClick,
             icon = Icons.Outlined.Settings,
             text = stringResource(R.string.settings),
+            showExpertBadge = isExpertModeEnabled,
             modifier = Modifier.weight(1f)
         )
     }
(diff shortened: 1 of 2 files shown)


MarcaDian:
I didn't test if it works completely as it should.

LisoUseInAIKyrios:
I will try later with a mock third party patch bundle for YT that adds 2 additional patches.

Update: Multiple patch bundles work.

MarcaDian:
Damn, that still didn't fix it.

https://github.com/user-attachments/assets/9d809056-476a-473e-848c-4993b78792f4

LisoUseInAIKyrios:
1. If there's only one bundle (the built in bundle), it could be pre-expanded to save a tap.
2. The rename function could be hidden for built in bundles.
3. Should "Patch bundles" be renamed to "Patch sources"?

MarcaDian:
> * If there's only one bundle (the built in bundle), it could be pre-expanded to save a tap.
> * The rename function could be hidden for built in bundles.
> * Should "Patch bundles" be renamed to "Patch sources"?

This is still a "draft", since a lot of things had to be written from scratch. So all this can be done when we test that everything is fine.
Honestly, I don't want this PR anymore)) Why break something that already works. I don't really like the design of this solution right now, since it should be universal.

LisoUseInAIKyrios:
I think this design looks ok. This change is needed along with the last missing use case of updating existing patched app, and then we can hide Expert mode entirely.

We can still hide away all this complexity by changing "Expert" mode to a switch, that when enabled it exposes all the stuff the average user will never use:
- Patch sources
- Import / Export keystore
- Disable patcher process & process memory limit
- Enable / Disable individual patches (can be a step in the patching process). Placing patch options here is a possibility and then options are not anywhere else.


When expert mode is off, the existing Morphe patch options could still show in the settings menu. Basically instead of a completely different UI we change Morphe UI to a two state design: Simple / Expert.

LisoUseInAIKyrios:
And I don't think many users bother with the patch options. So if they are always in the patch sources that's also ok. If this is done along with an "Expert" switch, then "Expert exposes the controls to add additional patch sources.

LisoUseInAIKyrios:
Multiple bundles patch correctly. Created a project with a single YT patch, added it as a source, and simple patching applied all Morphe patches and the additional patch in the custom bundle.

MarcaDian:
> We can still hide away all this complexity by changing "Expert" mode to a switch

I was also thinking about this, just implementing it through the settings like we are doing now through checking Morpher mode in some places.

> The weird scroll infinite bounce issue seems to show up in the patch option screen.

The internet says that for complex menus like ours, we should use `BottomSheetScaffold` instead of `ModalBottomSheet` and that should fix the problem.

LisoUseInAIKyrios:
Keeping the patch options in the settings is ok. 

There is one last use case I forgot: Including / excluding patches. This is needed for most 3rd party patches because they almost never have in-app patch settings. This can be exposed with an "Expert" switch, and picking which patches is a step in the patcher flow just before patching. The patch options could be shown here if we wanted, and it would make simple mode ever simpler and hide away patch options that are rarely used.

MarcaDian:
Found a problem where the music gets a light theme key that is not in the patch.

```
app.morphe.patcher.patch.OptionException$OptionNotFoundException: No option with key lightThemeBackgroundColor
	at app.morphe.patcher.patch.Options.get(Option.kt:179)
	at app.morphe.patcher.patch.Options.get(Option.kt:145)
	at app.morphe.patcher.patch.Options.set(Option.kt:159)
	at app.revanced.manager.patcher.runtime.CoroutineRuntime.execute(CoroutineRuntime.kt:53)
	at app.revanced.manager.patcher.worker.PatcherWorker.runPatcher(PatcherWorker.kt:266)
	at app.revanced.manager.patcher.worker.PatcherWorker.access$runPatcher(PatcherWorker.kt:62)
	at app.revanced.manager.patcher.worker.PatcherWorker$runPatcher$1.invokeSuspend(Unknown Source:15)
	at kotlin.coroutines.jvm.internal.BaseContinuationImpl.resumeWith(ContinuationImpl.kt:33)
	at kotlinx.coroutines.DispatchedTask.run(DispatchedTask.kt:100)
	at kotlinx.coroutines.scheduling.CoroutineScheduler.runSafely(CoroutineScheduler.kt:586)
	at kotlinx.coroutines.scheduling.CoroutineScheduler$Worker.executeTask(CoroutineScheduler.kt:829)
	at kotlinx.coroutines.scheduling.CoroutineScheduler$Worker.runWorker(CoroutineScheduler.kt:717)
	at kotlinx.coroutines.scheduling.CoroutineScheduler$Worker.run(CoroutineScheduler.kt:704)

```

LisoUseInAIKyrios:
Yeah YT Music doesn't have a light mode (it's always dark), so it doesn't have a light theme patch option.

MarcaDian:
I think the basic design is ready) Does it work? Not sure 😅.

Again, there is no electricity to complete everything as planned.

But at first glance it seems to work.

MarcaDian:
<details>








</details>

LisoUseInAIKyrios:
This is awesome.

I love the expert mode gear head switch icon.

1. If other patch bundles exists, the home page needs a way to select a different apk. It could be a "Other app" button that was present in early pre-release versions.
2. The include/exclude selector for the patch options might work better if it doesn't show an X icon or if the include/exclude state is shown differently. Because at first glance it's unclear if the X means it's not currently included or if X is an action. Maybe show a checkmark if currently included and no icon or an empty circle or something when not included.
3. If only one patch bundle exists (the built in Morphe patches), the bundle disable button could not be shown because disabling means no patch sources.
4. Not related to this PR, but the patch bundle patch list and changelog buttons could have outlined borders or a change to make the buttons look more clickable.

LisoUseInAIKyrios:
Not required for this PR, but if the last use cases are handled by Morphe layer then the old expert mode can be completely removed as obsolete:

1. Handle prompting/updating existing patched apps.
2. Add patcher process switch and memory setting
3. Add remove unused architectures setting

Lots of different ways to handle item 1. It could be a prompt to update, or a separate screen of installed apps. Or the home page could be slightly changed to both show existing installed apps and a way to add patched apps.

For 2 and 3, maybe only show if expert mode is enabled, and if this is done then also only show import/export keystore when expert is enabled.

MarcaDian:
🔥


<details>
<summary>Screenshot</summary>



</details>

MarcaDian:
Having some free time I tested this PR, and it works great.

**Simple mode**, if the patches options are selected from the settings, they are applied.

**Expert mode**, ignores the simple mode options in the settings.

That is, the patches options isolation works.

MarcaDian:
To add a separate screen for Installed/Patched apps, we probably need to change the concept of the main screen.

Honestly, remember, you suggested adding just 2 buttons for patching. And now we are expanding the functionality, and that is already not enough.

Maybe instead of 2(3) floating buttons, implement something like a button panel at the bottom of the page? I can't come to the right solution yet.

LisoUseInAIKyrios:
Yes it gets more complicated when updates and other apps are brought in.

MarcaDian:
<details>



</details>

MarcaDian:
Also, if I understand correctly, we need to remove the `Fallback installer` option?

LisoUseInAIKyrios:
Yes it can be hidden or disabled entirely

app/src/main/java/app/revanced/manager/ui/component/morphe/home/HomeBottomActionBar.kt
```diff
@@ -0,0 +1,139 @@
...
+fun HomeBottomActionBar(
+    onInstalledAppsClick: () -> Unit,
+    onBundlesClick: () -> Unit,
+    onSettingsClick: () -> Unit,
+    modifier: Modifier = Modifier
+) {
+    Row(
+        modifier = modifier
+            .fillMaxWidth()
+            .padding(vertical = 24.dp, horizontal = 16.dp),
```
wchill: All of these constants should really be named constants, for UI consistency. Don't necessarily need to fix this problem for all the constants in the entire project right away, but it would be a good idea to at least have the new UI elements use them.

MarcaDian:
I'm having some problems adapting the update screen from legacy code. Namely, the Repatch button opens the old menus for repatching, but they don't have the code to preserve the original APK.
Maybe for now we can open the same dialogs as on the main page.

<details>






</details>

LisoUseInAIKyrios:
Yes the home screen patcher flow could be used to update the app. Optionally the patching flow UI text could display "update" language instead of "install". Reusing the original unpatched APK can be added later.

MarcaDian:
Even though the Repatch button isn't working yet, it's ready for review. I'm just tired of this PR, it's too complicated for me.

app/src/main/java/app/revanced/manager/ui/component/morphe/home/HomeBundleDialogs.kt
```diff
@@ -0,0 +1,326 @@
...
+ * Dialog for adding patch bundles
+ */
+@Composable
+fun MorpheAddBundleDialog(
+    onDismiss: () -> Unit,
+    onLocalSubmit: () -> Unit,
+    onRemoteSubmit: (url: String) -> Unit,
+    onLocalPick: () -> Unit,
+    selectedLocalPath: String?
+) {
```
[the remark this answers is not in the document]
LisoUseInAIKyrios: Still shows zero patches and I can't begin patching. Tried clearing app data and with/without pre-release. I can help figure out what's wrong sometime tomorrow, if its still doing this.

Morphe

app/src/main/java/app/revanced/manager/ui/screen/MorpheHomeScreen.kt
```diff
@@ -167,119 +131,66 @@ fun MorpheHomeScreen(
...
-                        contentDescription = stringResource(R.string.morphe_home_bundles),
-                        containerColor = MaterialTheme.colorScheme.secondaryContainer,
-                        contentColor = MaterialTheme.colorScheme.onSecondaryContainer
-                    )
+            // Other apps button (only show in expert mode)
+            showOtherAppsButton = useExpertMode,
+            onOtherAppsClick = {
+                if (availablePatches <= 0 || bundleUpdateProgress != null) {
+                    context.toast(context.getString(R.string.morphe_home_patches_are_loading))
+                    return@HomeSectionsLayout
```
LisoUseInAIKyrios: If a third party patch bundle exists, then maybe "other apps" should be shown in simple mode? It can patch using the same logic as patching YT in simple mode, and use the recommended patches without asking which to include/exclude.

app/src/main/java/app/revanced/manager/ui/component/morphe/settings/OriginalApksManagementDialog.kt
```diff
@@ -0,0 +1,244 @@
...
+import app.revanced.manager.domain.repository.OriginalApkRepository
+import app.revanced.manager.ui.component.AppIcon
+import app.revanced.manager.ui.component.morphe.shared.*
+import app.revanced.manager.ui.component.morphe.utils.formatBytes
+import app.revanced.manager.util.toast
+import kotlinx.coroutines.launch
+import org.koin.compose.koinInject
+
+@Composable
+fun OriginalApksManagementDialog(
```
LisoUseInAIKyrios: Maybe in the future this can be a single expert button of "Remove unused original apks", and it deletes all original apk's that do not have a patched version currently installed. Then there's no UI submenu.

Or maybe automatically delete all unused apk's and then there's no menus at all. Original APK's could be deleted about 1 day/week after all patched versions of the app are uninstalled. Not sure what way to check if enough time has gone by to delete, but maybe a "first unused" nullable field would work and on app startup or some other time it checks what apps are now uninstalled and set/update the field. If "first unused" is over a day/week old then remove the original apk.

MarcaDian: In the current logic, the old original APK is deleted when a new version is patched.

MarcaDian:
I think this can be done endlessly)

LisoUseInAIKyrios:
Yes. When this reaches a point of "done for now", we should merge.

LisoUseInAIKyrios:
Pixel 2 emulator (Android 10 was used), the bottom buttons are clipped slightly. May be something with the on screen nav bar.

<details>

pixel 2
</details>

app/src/main/java/app/revanced/manager/ui/component/morphe/home/HomeInstalledAppsScreen.kt
```diff
@@ -0,0 +1,199 @@
...
+import app.revanced.manager.ui.component.morphe.shared.MorpheCard
+import app.revanced.manager.ui.component.morphe.shared.SectionTitle
+import app.revanced.manager.ui.viewmodel.InstalledAppsViewModel
+import org.koin.androidx.compose.koinViewModel
+
+@Composable
+fun HomeInstalledAppsScreen(
+    onAppClick: (InstalledApp) -> Unit,
+    viewModel: InstalledAppsViewModel = koinViewModel()
+) {
```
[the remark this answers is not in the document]
LisoUseInAIKyrios: Before merging, should try installing the last stable/dev release, patch and install an app, then install built manager from this PR, and verify the installed app appears/works correctly in Manager.

MarcaDian:
Hm...This should take into account system bars. 

https://github.com/MorpheApp/morphe-manager/blob/c6d98f630f1c3155b913711bfd49ef8bb48d00ac/app/src/main/java/app/revanced/manager/ui/screen/MorpheHomeScreen.kt#L244

https://github.com/MorpheApp/morphe-manager/blob/c6d98f630f1c3155b913711bfd49ef8bb48d00ac/app/src/main/java/app/revanced/manager/ui/component/morphe/home/HomeSectionsLayout.kt#L95

It may be that the elements no longer have space on the screen.

ArcherLegend:
Could you add the possibility to change the package name with the app on Android?
Do you plan to have a GUI for Win-Mac-Lin like inotia00's rvx-builder?

LisoUseInAIKyrios:
Package name can already be changed with the universal change package name patch. That is unrelated to this PR.

LisoUseInAIKyrios:
@MarcaDian Ready to merge?

MarcaDian:
Wait a bit :-)

MarcaDian:
Probably ready 🤞

LisoUseInAIKyrios:
Ok squash ahead

PR #134, opened 2026-01-29
fix: 'GmsCore support' patch not excluded in root mode

MarcaDian:


MarcaDian:
I also wanted to change the logic of app selection if the default bundle is disabled so that it searches for available bundles and patches, ~but for some reason now the app crashes...~
This was due to the remnants of another PR, clearing the application data helped

MarcaDian:
I don't know if I set correct behavior:

In simple mode:
- the main buttons work only with the default bundle.
- the Other apps button searches for available patches in the default bundle, if not found then in other bundles.
- if the default bundle is disabled, then when selecting the main buttons there will be a toast asking to enable it.

In expert mode without restrictions.

LisoUseInAIKyrios:
I think all of that seems correct.

The second item should fix users who enter RVX patch repos but forget to disable built in patches, then it fails when basically double patches are attempted.

LisoUseInAIKyrios:
I checked and root works (GmsCore is excluded by default).

LisoUseInAIKyrios:
Looks good from here. Merge whenever.

PR #137, opened 2026-01-30
feat: Use APKEditor for APKM merging

LisoUseInAIKyrios:
Pulls in fixes by the awesome [Jman](https://github.com/Jman-Github/Universal-ReVanced-Manager).

Changes split apkm merging from [ARSCLib](https://github.com/REAndroid/ARSCLib) to [APKEditor](https://github.com/REAndroid/APKEditor)

### DIFF
--- a/app/src/main/java/app/morphe/manager/patcher/split/Merger.kt
+++ b/app/src/main/java/app/morphe/manager/patcher/split/Merger.kt
@@ -6,7 +6,6 @@ import com.reandroid.apk.ApkBundle
 import com.reandroid.apk.ApkModule
 import com.reandroid.app.AndroidManifest
 import com.reandroid.archive.ZipEntryMap
-import com.reandroid.archive.block.ApkSignatureBlock
 import com.reandroid.arsc.chunk.xml.ResXmlElement
 import com.reandroid.arsc.container.SpecTypePair
 import com.reandroid.arsc.header.TableHeader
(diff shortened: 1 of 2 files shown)


optimatrix0:
Hello, good morning.

I’m posting here again, but instead of opening a new issue, I preferred to contribute to an ongoing piece of work.

Something unexpected happened that affected me directly. I was using the debug version from the “remove dead code” branch to test accessibility improvements. However, when I updated to version 1.4.0-dev.2, which already includes those merged changes, the app stopped opening entirely.

I wanted to report this here because this is the only contribution I can realistically make in this situation, and it was a noticeable inconvenience. I understand this may be a small issue, and that I might need to reinstall everything to test whether it works again, but I still felt it was important to mention what happened.

Even though this was a debug build, this kind of crash ideally shouldn’t happen, especially since the project technically supports moving between debug, development, and stable builds through timestamp-based versioning.

What I want to say is that this was inconvenient, and I hope this comment helps in some way. It actually reminded me of a similar issue that happened in the previous project, where a similar problem took about five days to be fixed. In this case, I believe it will be resolved much faster.

That’s all. I hope the community receives this small contribution positively. And if there’s any issue with me posting this here, you can feel free to give me a verbal warning — I’ll take it without any problem.

Jman-Github:
> Using APKEditor.jar command line works, but it's not yet clear what command line is doing differently that needs to be applied here.

@LisoUseInAIKyrios

I think the cause here is that URV is using `ApkModule.merge(..., mergeResources = false)` while the APKEditor JAR uses `ApkBundle.mergeModules()`. The `ApkBundle.mergeModules()` way merges resource tables and remaps resource IDs across modules much more effectively than the `ApkModule.merge(..., mergeResources = false)`. The reason why URV was using `ApkModule.merge(..., mergeResources = false)` in the first place is because I needed to see progress on each split so the patcher screen can show progress for them. But I can use ApkLogger to create custom log outputs for each split, and use `ApkBundle.mergeModules()` instead just fine. So I'm thinking these inconsistencies will be fixed after I do this.

Jman-Github:
This issue is now resolved as of the [latest commit](https://github.com/Jman-Github/Universal-ReVanced-Manager/pull/136/changes/8c97bd84458232159c7ba636708b63b629064102) for the URV Manager. I tested it by patching the split APK of MyFitnessPal (the app that was reported to be having problems when patched and merged with either of our managers), which was successful and resulted in zero crashes. I made the merger follow the same process ApkEditor does exactly. The only parts that aren't part of ApkEditor are the native libs stripper and unused split remover, which are toggles in URV. Those were tested too, and don't result in any errors/crashes. There is also additional logging, but that doesn't interfere with merging at all and was added, as stated previously, for URV's patcher progress UI.

LisoUseInAIKyrios:
Everything seems to be working now. Can patch MyFitnessPal APKM bundle and it no longer crashes on startup.

PR #138, opened 2026-01-31
feat(Custom branding): Allow Manager to process custom icon/headers into the correct formats/names/sizes

MarcaDian:
This PR introduces two simple dialogs: one for creating an adaptive icon for a branding patch, and another for creating a header for a header patch.

In the preview, images can be moved and scaled using gestures. The saved output will exactly match what the user sees in the preview.

For convenience, snapping to the center of the preview has been added.

When saving the result, the user must select a directory. Inside it, a `morphe_branding` folder will be created.

Within this folder:
- `morphe_icon` will be created for the generated icon
- `morphe_header` will be created for the generated header

Each of these will contain a single `density-xxxx` folder that corresponds to the user’s current screen density.
___


MarcaDian:
This PR depends on https://github.com/MorpheApp/morphe-manager/pull/132, as it was created on its basis.

MarcaDian:
The only thing I've noticed, need to apply anti-aliasing to png when saving.

LisoUseInAIKyrios:
It may need bicubic or some other anti-aliasing during the rescaling.

https://www.w3tutorials.net/blog/draw-smoothly-scaled-bitmaps-on-canvas

MarcaDian:
I wonder why there are conflicts, since the base branch hasn't changed?

LisoUseInAIKyrios:
Changes to dev?

MarcaDian:
This PR was based on `remove-dead-code` branch, which was merged with `dev` branch, so there shouldn't be any conflicts?

> Base automatically changed from remove-dead-code to dev [17 minutes ago](https://github.com/MorpheApp/morphe-manager/pull/138#event-22441725543)

LisoUseInAIKyrios:
It's because the strings.xml was changed both on dev and here. I think it's resolved now.

MarcaDian:
It's strange that all the commits from the previous branch are showing.

MarcaDian:
I want to add support for the Update gesture on the home screen, so I want to make sure everything is ok).
This should probably be done in a separate PR.

LisoUseInAIKyrios:
Old commits are because this branch was forked from the cleanup branch. Can ignore because the changes shown are just what's here.

MarcaDian:
Can merge this?

PR #147, opened 2026-02-01
fix: UX improvements

MarcaDian:
Changes:
- Now clicking on a patch card activates/deactivates a patch instaed of individual buttons.
- Added a warning dialog if patches from multiple bundles are selected.
- Redesigned the appearance screen to use cards instead of horizontal scrolling.

___


LandonMoran:
Did you make sure that this works us talk back properly? If not I can make sure of that

MarcaDian:
Can merge this?

LisoUseInAIKyrios:
Yes I think so

PR #155, opened 2026-02-03
feat: Add an extended list of supported versions

MarcaDian:
--- 

---

### DIFF
--- a/app/src/main/java/app/revanced/manager/ui/screen/home/ExpertModeDialog.kt
+++ b/app/src/main/java/app/revanced/manager/ui/screen/home/ExpertModeDialog.kt
@@ -936,7 +936,7 @@ private fun PathInputOption(
         // Create Icon button
         if (isIconField) {
             MorpheDialogOutlinedButton(
-                text = stringResource(R.string.adaptive_icon_create_new),
+                text = stringResource(R.string.adaptive_icon_create),
                 onClick = { showIconCreator = true },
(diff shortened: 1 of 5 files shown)


optimatrix0:
If you allow me to make a suggestion regarding this feature that’s being implemented: it would be very useful if, when users are in advanced mode, experimental compatibility (YouTube, etc.) could also be shown.
Of course, always with clear warnings and alerts, and with stable versions being the primary recommendation.

This could be presented in a separate list or even in a dedicated section specifically for experimental versions. That would help a lot, since having to manually hunt for the app on the website can be inconvenient for many users and feels very similar to the experience of older projects.

This approach would make things easier for new users as well and would add a nice new feature overall.
As an alternative (or in addition), the website could also include a regularly updated section listing supported versions, so users can easily access them when needed.

Either way, thank you for your time and consideration.

MarcaDian:
In the future, it is planned to add the "Experimental" tag to the list so that users can see that this is an experimental version that is still in development and may have bugs and side effects.
But this requires support from the patcher and patches first, and then the manager.

FaiyazAhamadRabbi:
Exactly we need this.

FaiyazAhamadRabbi:
Still we missed some usefull features in latest version of Morphe manager. Like Patches new update dialog. In the new version which bugs fixed and which new features adds. Patch updated automatically so for it need a toogle switch .Plz add this nex update(sorry for my bad English).

PR #161, opened 2026-02-04
feat: Add stored patch selection dialog

MarcaDian:
Depends on https://github.com/MorpheApp/morphe-manager/pull/157

### Scenario 1: Normal Patching → Installation
Patching → Installation → Save to Database:
- InstalledApp (app record)
- AppliedPatch (list of patches)
- SelectionPayload (detailed selection info)
- Selection (saved patch choices - persists after uninstall)
- Options (patch configuration - persists after uninstall)

### Scenario 2: Delete via App Info Dialog
  InstalledAppInfoDialog → removeAppCompletely():
- Deletes InstalledApp record
- Deletes AppliedPatch records
- Deletes patched APK file
- Deletes original APK file
- Does NOT delete Selection (preserved in PatchSelectionRepository)
- Does NOT delete Options (preserved in PatchOptionsRepository)

### Scenario 3: Delete via Storage Management
  Settings → Storage Management → Patched APKs → Delete:
- Deletes InstalledApp record
- Deletes patched APK file
- Deletes original APK file (if using removeAppCompletely)
- Does NOT delete Selection
- Does NOT delete Options

### Scenario 4: Delete via Patch Selections Management
  Settings → Storage Management → Patch Selections → Delete:
- Deletes Selection for selected package
- Deletes Options for selected package
- Does NOT affect InstalledApp records
- Does NOT affect APK files

### Scenario 5: Re-patching After Deletion
  After deleting app via Scenario 2 or 3:
1. User selects same package to patch
2. System automatically loads preserved Selection
3. System automatically loads preserved Options
4. User patches with previous configuration instantly applied
5. No need to manually select patches again
___


MarcaDian:
Patch selection and patch options are not deleted from the database, the only way to delete them is manually in the settings `Storage Management - Patch selections`

PR #170, opened 2026-02-06
feat: Add installer prompt on patcher screen

MarcaDian:
## How It Works

- Settings dialog and Patcher dialog use the same `InstallerSelectionDialog` component
- Both dialogs update installer availability every 1.5 seconds via LaunchedEffect
- When toggle is enabled, `Install` button shows selection dialog
- Selected installer is stored as `oneTimeInstallerToken`
- Installation temporarily swaps primary token, resolves plan, then restores original
- One-time token is cleared after use

### DIFF
--- a/app/src/main/java/app/revanced/manager/ui/screen/settings/system/InstallerSection.kt
+++ b/app/src/main/java/app/revanced/manager/ui/screen/settings/system/InstallerSection.kt
@@ -43,6 +43,7 @@ fun InstallerSection(
     onShowInstallerDialog: () -> Unit
 ) {
     val coroutineScope = rememberCoroutineScope()
+    val expertMode by settingsViewModel.prefs.useExpertMode.getAsState()
     val primaryPreference by settingsViewModel.prefs.installerPrimary.getAsState()
     val primaryToken = remember(primaryPreference) {
         installerManager.parseToken(primaryPreference)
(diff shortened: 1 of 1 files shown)


LisoUseInAIKyrios:
Is an expert mode feature?

MarcaDian:
> Is an expert mode feature?

Should this only be in expert mode??

LisoUseInAIKyrios:
It's a setting to prompt on each install? It can be either regular or expert I guess.

MarcaDian:
> It's a setting to prompt on each install?

Yes. I won't be using this anyway, so I don't know whether to do this just for expert mode or in all modes.
Now it's for all modes.

LisoUseInAIKyrios:
Then can show the setting in expert mode.

If a user enabled in expert mode, turned on the setting, changed to normal mode, then the "choose installer" prompt can still be shown.

bikram-agarwal:
Installed Morphe Manager 1.6.0. Enabled "prompt installer on install". Still only getting mount option on patching screen. For example, this is at the end of patching Reddit - 
Screenshot_20260208-102833

MarcaDian:
@bikram-agarwal I haven't tested root mode because I don't have it. What happens if click the Mount button?

bikram-agarwal:
> @bikram-agarwal I haven't tested root mode because I don't have it. What happens if click the Mount button?

It mounts in root mode. And that then shows up as module in Root Manager app too. (I have Wild KSU). 

I was expecting that there'll be an "Install" button here, and tapping that will prompt me to choose a method - Root Mount or Stock Package Manager or any other package manager the device has.

LisoUseInAIKyrios:
To install with mount, GmsCore must be excluded. So the installation method needs to be known before patching even starts. And even if selecting patches, GmsCore is force included/excluded based on installation mode.

MarcaDian:
@bikram-agarwal Try this artifact. The pre-patching installer type dialog is called before patching. I think it's better than nothing.

https://github.com/MorpheApp/morphe-manager/actions/runs/21832547342/artifacts/5435086165

> Dialog shown to root device users before patching to choose between Root Mount and Standard Install.
> 
> The installation method directly affects how the APK is patched:
> - **Root Mount** excludes the GmsCore support patch - the mounted APK replaces the stock APK in-place via bind-mount, so the original Google services remain available and GmsCore would actually interfere.
>  - **Standard Install** includes GmsCore support - the patched APK is installed as a separate app that needs the microG / GmsCore bridge to communicate with Google services.
> 
> Because of this, the choice cannot be deferred to after patching. The user must decide now so the correct set of patches is applied.

bikram-agarwal:
Yep, this seems to be working. It asked me install method beforehand and then again at the end. So, I was able to use package manager for Reddit patching, even though root mounting is my default. Thanks for this. 

Question though - 

GmsCore support is needed for only Google apps like YT / YTM, right? Reddit, for example, doesn't have / use that. So, can't we only show install option at the end?

MarcaDian:
@bikram-agarwal please test this artifact
https://github.com/MorpheApp/morphe-manager/actions/runs/21868592271/artifacts/5449739721

bikram-agarwal:
> @bikram-agarwal please test this artifact https://github.com/MorpheApp/morphe-manager/actions/runs/21868592271/artifacts/5449739721

Worked as expected for Reddit. Asked me to pick installer only after patching and I was able to install using package manager. 

Everything was going as expected for YTM too. Asked me to pick installer only after patching and I picked the root mount. That's when it threw error saying "_Package conflict_", even though the existing version was also patched with Morphe. I tap uninstall on the "_uninstall the existing version_" prompt. It then says "your app is patched and installed" but there's still a red ❌ there, and YTM is not actually patched. It's now at stock factory image version of YTM.
Morphe YTM

MarcaDian:
Before your answer, I tried to test with a "fake" root mode on my device and noticed that sometimes the root mode flag was set incorrectly or untimely, because of this the `Mount`/`Install` button and some root mode badges did not work correctly.

Since I did not make the root mount code, it is difficult for me to answer and understand the correct operation of the root mount.

In my tests, I did not have the same behavior as you, although I tried about 20 different patching and mounting methods.

Theoretically, I laid the following logic:

- if the root mode works correctly and the root mount is selected in the settings, then when starting the patching, a dialog appears asking you to choose the installer: root or regular.
- If you choose regular, the root mode is temporarily disabled, the GmsCore patch is active, and at the end of the patching, the behavior will be the same as with non-root.
- If you choose the root mode, the GmsCore patch is inactive and at the end of the patching, the mount will be performed.

bikram-agarwal:
fwiw, the previous test artifact you had shared here - the one that showed "choose installer method" both before and after patching - that is working fine for both reddit and YTM.

PR #179, opened 2026-02-08
fix: File system access request appeared where it wasn't needed

MarcaDian:


MarcaDian:
I've verified that it works for me.
Now the file system access request is only appear when selecting or creating icons for the branding/header patches.

LisoUseInAIKyrios (approved):

PR #186, opened 2026-02-09
feat: Get manager release info from static JSON file

MarcaDian:


LandonMoran:
What exactly is this for?

LisoUseInAIKyrios:
It can fix manager not downloading patches and other network errors if the user IP is blocked from GitHub API.

MarcaDian:
I don't know if this will work, Android Studio didn't allow me to rename it with the built-in tools, so I did it manually.

Also tried adding migration code for the icon activity.

Also for the root module I don't know if migration code will work.

LisoUseInAIKyrios:
I checked now and root works.

MarcaDian:
I checked the icon migration activity, but the previously selected icon was not selected. But the application did not disappear from the launcher, so we can assume that it works.

LandonMoran:
This did not fix my issue

LisoUseInAIKyrios:
Then it may be something with your internet connection and connecting to GitHub itself.

LandonMoran:
Yes, I think it might be but I'm not the only one this happens to a VPN will fix it right? So that also makes sense. But then why would my internet be blocking it? Because in the old revanced I could do it which also reaches out to the GitHub API> Then it may be something with your internet connection and connecting to GitHub itself.

LisoUseInAIKyrios:
If it is something with your internet connection or it's GitHub blocking your ip, then a VPN should fix it.

LandonMoran:
Yes, and a VPN does fix it, but my main question is why does this happen? Because I didn't have that issue with revanced or rvx

LandonMoran:
Since it works in other projects and can connect to the GitHub API then my IP isn't blocked but I am not exactly sure

LisoUseInAIKyrios:
Maybe it's something new. RVX uses GitHub too, so maybe it's a coincidence it's starting now.

PR #197, opened 2026-02-13
feat: New patch selections dialog

MarcaDian:


LisoUseInAIKyrios:
Is this ready to merge?

MarcaDian:
No still in progress

MarcaDian:
All this time I was fighting windmills (Don Quixote), I don't know for what reason the database kept the selections of patches and options for both packages: original and patched. I don't remember if it was upstream or my bad refactorings. But finally I found the reason that all this time I was using crutches in different parts of the code to filter by the original package.

PR #217, opened 2026-02-17
feat: Show Android notifications when patch and manager updates are available

MarcaDian:
## Overview

WorkManager with `PeriodicWorkRequest` is unreliable for timely update notifications. On Android 6+ devices, Doze mode delays background tasks by hours. On Xiaomi, Samsung, Huawei, and OnePlus devices, aggressive battery optimisation systems kill background processes entirely. In practice, users often never receive notifications unless they open the app manually.

## What changed

**Firebase Cloud Messaging (FCM)** is now the primary notification delivery channel. High-priority FCM messages are delivered by Google Play Services - a system process exempt from Doze restrictions - ensuring notifications arrive even on devices with aggressive battery optimisation. WorkManager remains as a silent fallback for devices without Google Play Services.

### Four-topic subscription model

Manager and patch notifications are independent, each with a stable and dev variant:

| Topic | Audience |
|---|---|
| `morphe_updates` | Manager updates - stable build with prereleases OFF |
| `morphe_updates_dev` | Manager updates - dev build OR prereleases ON |
| `morphe_patches_updates` | Patch updates - prereleases OFF |
| `morphe_patches_updates_dev` | Patch updates - prereleases ON |

A device running a dev manager build with prereleases OFF subscribes to **both** manager topics, since a stable release (e.g. `1.5.0`) is a valid upgrade from a dev build (e.g. `1.5.0-dev.1`). The patches topic is determined solely by the "Use prereleases" preference, independent of the manager build variant.

### Settings UI

On devices with Google Play Services, the WorkManager check frequency selector is hidden - it is not relevant when FCM handles delivery. Devices without GMS retain the full interval configuration UI.

### Release workflow

FCM notifications are sent automatically from GitHub Actions after each release.

### DIFF
--- a/app/google-services.json
+++ b/app/google-services.json
@@ -7,7 +7,7 @@
   "client": [
     {
       "client_info": {
-        "mobilesdk_app_id": "1:574971415597:android:baeaf1958ebb5a84dc7ca4",
+        "mobilesdk_app_id": "1:574971415597:android:5acf99d4cc8260eedc7ca4",
         "android_client_info": {
           "package_name": "app.morphe.manager"
         }
@@ -15,7 +15,7 @@
       "oauth_client": [],
       "api_key": [
         {
-          "current_key": "AIzaSyCq_Jp7UkUXovVxBJBb8WMFfCY29q_pBXA"
+          "current_key": "AIzaSyAICrE-tnEhNKXh1vEbfucYWWxOr7NSr_g"
         }
       ],
       "services": {


MarcaDian:
Interestingly, with this PR, I started getting patching notifications. I've never seen them before, even upstream. Probably because permission for notification was not given.

Jman-Github:
Couldn't we use something like this? There are community servers hosted too so you don't have to necessarily self host one.
https://unifiedpush.org/

LisoUseInAIKyrios:
A push notification service would be best. 

We can later figure out something.

MarcaDian:
I have a Xiaomi and notifications still don't come when patches are released. I'll try disabling battery optimization for the manager.

optimatrix0:
Would it be possible to add a bubble-style notification for important updates? For example, when a new Manager version is released, a floating bubble could appear (similar to Google Messages chat heads), ensuring the user actually notices it, with a clear “Dismiss” button above it.

That would be really helpful, since critical updates could no longer be missed, especially by users who don’t regularly check the app or standard notifications. It could greatly improve visibility without requiring users to manually look for updates.

MarcaDian:
Ready to merge?)

LisoUseInAIKyrios:
Yes I think so

PR #231, opened 2026-02-21
feat: Use Morphe patcher 1.2.0

LisoUseInAIKyrios:
Changes to patcher 1.2.0 which internally uses ARSCLib for resource processing.

This now allows resource patching apps that previously failed, such as disabling play store updates (change version code) for Reddit.

### DIFF
--- a/.github/scripts/send_fcm.py
+++ b/.github/scripts/send_fcm.py
@@ -1,3 +1,6 @@
+# Copyright 2026 Morphe.
+# https://github.com/MorpheApp/morphe-manager
+
 #!/usr/bin/env python3
 """
 Send an FCM push notification to a Morphe Manager topic.
--- a/.github/workflows/build_pull_request.yml
+++ b/.github/workflows/build_pull_request.yml
@@ -34,7 +34,7 @@ jobs:
           ./gradlew assembleRelease --stacktrace
 
       - name: Upload APK
-        uses: actions/upload-artifact@v4
+        uses: actions/upload-artifact@v7
         with:
-          name: Morphe
+          archive: false
           path: app/build/outputs/apk/release/*.apk
--- a/app/CHANGELOG.md
+++ b/app/CHANGELOG.md
@@ -1,3 +1,38 @@
+## app [1.10.2](https://github.com/MorpheApp/morphe-manager/compare/v1.10.1...v1.10.2) (2026-03-02)
+
+
+### Bug Fixes
+
+* Manager does not show updates if available ([#270](https://github.com/MorpheApp/morphe-manager/issues/270)) ([4e6f2af](https://github.com/MorpheApp/morphe-manager/commit/4e6f2afee34894c271903b6ea18a2b1a2cfe5ee1))
+
+## app [1.10.2-dev.1](https://github.com/MorpheApp/morphe-manager/compare/v1.10.1...v1.10.2-dev.1) (2026-03-02)
+
+
+### Bug Fixes
+
+* Manager does not show updates if available ([#270](https://github.com/MorpheApp/morphe-manager/issues/270)) ([4e6f2af](https://github.com/MorpheApp/morphe-manager/commit/4e6f2afee34894c271903b6ea18a2b1a2cfe5ee1))
+
+## app [1.10.1](https://github.com/MorpheApp/morphe-manager/compare/v1.10.0...v1.10.1) (2026-03-02)
+
+
+### Bug Fixes
+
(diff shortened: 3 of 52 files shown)


LisoUseInAIKyrios:
Patch source screen still crashes. It's not proguard (I should have checked the logs and not assumed it was).

```
java.lang.NullPointerException
at io.ktor.util.TextKt.painterResource(Unknown Source:121)
at app.morphe.manager.ui.screen.home.SourceManagementSheetKt$BundleIcon$2.invoke(Unknown Source:858)
at androidx.compose.runtime.internal.ComposableLambdaImpl.invoke(SourceFile:25)
at androidx.compose.runtime.internal.ComposableLambdaImpl.invoke(SourceFile:1)
at androidx.compose.material3.SurfaceKt$$ExternalSyntheticLambda0.invoke(Unknown Source:185)
at androidx.compose.runtime.internal.ComposableLambdaImpl.invoke(SourceFile:25)
at androidx.compose.runtime.internal.ComposableLambdaImpl.invoke(SourceFile:1)
at androidx.compose.runtime.Updater.CompositionLocalProvider(SourceFile:28)
at androidx.compose.material3.SurfaceKt.Surface-T9BRK9s(Unknown Source:105)
at coil.util.-FileSystems.BundleIcon(Unknown Source:428)
at coil.util.-FileSystems.BundleCardHeader(Unknown Source:229)
at app.morphe.manager.ui.screen.home.SourceManagementSheetKt$BundleManagementCard$2.invoke(Unknown Source:388)
                                                                                                    
```

MarcaDian:
I know, I'm testing through GitHub because it's faster than building locally, I'll write when I find a solution.

MarcaDian:
I understood why the error occurred, I will try to solve it radically.

LisoUseInAIKyrios:
Problem seems fixed. Release build doesn't crash anymore.

MarcaDian:
What solution fixed? I force pushed twice, is the solution with the proguard? Because the release was still being assembled when you wrote this

MarcaDian:
It seems that the last solution also works, which will fix the R8 confusing error.

LisoUseInAIKyrios:
```
                    painter = rememberDrawablePainter(
                        drawable = AppCompatResources.getDrawable(context, R.drawable.ic_launcher_foreground)
                    ),
```

That fixed it.

I think the error was not proguard stripping, but instead it inlined ("optimized") some code incorrectly.

MarcaDian:
By the way, this might fix https://github.com/MorpheApp/morphe-manager/issues/236, but I'm not sure

LisoUseInAIKyrios:
Maybe we should turn off proguard optimizations for Manager namespace.

The optimizations it does are somewhat questionable. I don't know if proguard optimziation is now relevant, but a few years ago it never showed any performance gains of substance because the JVM already does a lot of what it's trying to do with bytecode optimizations.

MarcaDian:
Wow, I forgot how much smoother the release version is. The dev version I'm testing constantly has some interface lag because it contains debug code.

And it shows a slightly faster build time.

LisoUseInAIKyrios:
Can add a second Android Studio run target to use release build settings (but not use proguard)

MarcaDian:
Logs with timing
[new_arclib.txt](https://github.com/user-attachments/files/25622557/new_arclib.txt)
[old_apktools.txt](https://github.com/user-attachments/files/25622559/old_apktools.txt)


image

LisoUseInAIKyrios:
The arsclib log doesn't include the preprocessing steps 

All arsclib post processing steps is ~22 seconds total. On my Android 1 year old flagship it is about 10 seconds total.

The post processing already had a bunch of changes to get reasonable patching times on Android (before it was up to 10 minutes). On Java desktop it's always very fast even before the latest changes.

```
2026-02-28 18:17:08.070 29921-30694 Morphe Patcher          app.morphe.manager                   I  Unescaping strings
2026-02-28 18:17:29.825 29921-30694 Morphe Patcher          app.morphe.manager                   I  Writing resource APK

MarcaDian:
So far, I don't see any objective reason not to merge this PR.

LisoUseInAIKyrios:
We could merge this with dev patcher/library dev dependencies.

Then later merge patcher/library/cli/manager to main all at the same time.

We should merge manager dev to main before merging this PR.

wchill:
If we want to merge this, imo we should either do a new stable release first or add a settings toggle for the tool to use. The patcher is stable enough on Linux CLI that I'm personally ok with the latter, even with the additional slowdown on Android (which we can work on resolving later, but will take time to do).

LisoUseInAIKyrios:
Will merge this in 2 days if manager stable release has no new urgent/easy bug fixes.

PR #232, opened 2026-02-21
feat: Show all patched apps on homescreen

MarcaDian:
## Changes

- Show all apps that have patches in all active bundles on the home screen
- Add long-press dialog to hide/unhide an app
- Add dialog button at the end of the list to view and manage hidden apps
- Refactor constants implementation
- Refactor bundle cards in the bottom sheet:
  - Add state change animations  
  - Expand displayed information for better user understanding
___

### DIFF
--- a/.github/workflows/build_pull_request.yml
+++ b/.github/workflows/build_pull_request.yml
@@ -31,7 +31,7 @@ jobs:
           KEYSTORE_ENTRY_PASSWORD: ${{ secrets.KEYSTORE_ENTRY_PASSWORD }}
         run: |
           chmod +x ./gradlew
-          ./gradlew assembleRelease --no-daemon --stacktrace
+          ./gradlew assembleRelease --stacktrace
 
       - name: Upload APK
         uses: actions/upload-artifact@v4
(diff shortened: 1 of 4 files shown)


MarcaDian:
I think it's ready for review and testing before merging.

Not4GoodBoyThou:
Although I did patched both of these before, I don't have both Proton VPN and Smart Launcher installed at the moment. Is it supposed to show like it?

Screenshot_20260223_143901_Morphe

MarcaDian:
Check if the patched APK is saved in the manager settings.

Not4GoodBoyThou:
It is. So does Solid Explorer and it shows uninstalled which is correct.

MarcaDian:
Try to force restart manager.

This does not always work when the manager is open at the time of uninstallation.

It makes no sense to complicate the logic of checking for application uninstallation while the manager is running, as this is a very unlikely case for a average user.

Not4GoodBoyThou:
> Try to force restart manager.

The things I did:
1. Force closed the manager. Issue not fixed.
2. Re added the patches after removing it. Issue not fixed.
3. Installed the saved apk, uninstalled within the manager itself. Issue not fixed.
4. Installed the saved apk, uninstalled from the device settings. Issue not fixed.

MarcaDian:
I just checked this case with Proton VPN. Can't reproduce issue.

1. I uninstalled patched Proton VPN, the manager did not show the status in the button, but it showed it in the button dialog.
2. Force restart the manager. The manager showed that the application was uninstalled.
3. I installed the patched one again. Force restart the manager, everything shows.

Try deleting the patched apk in the manager settings.

<details>




</details>

PR #247, opened 2026-02-25
feat: Enhance patch update management and mobile data controls

MarcaDian:
### Mobile Data Controls
- Add `Update on mobile data` toggle. Allows users to control whether the manager can update patches while using mobile/cellular data.
- Show confirmation dialog before patching on mobile data. When mobile data is detected, a warning dialog appears before starting the patching process.

### Patch Source Update Improvements
- Do not update disabled patch sources. Patch sources that are manually disabled will no longer be updated automatically.
- Improve update flow logic to better respect user-defined source states.

### Bundle Prerelease Refactor
- Split the global prerelease switch. Each bundle now has its own independent prerelease toggle.
- Improves flexibility and granular control over bundle versions.

### DIFF
--- a/app/src/main/java/app/morphe/manager/domain/repository/PatchBundleRepository.kt
+++ b/app/src/main/java/app/morphe/manager/domain/repository/PatchBundleRepository.kt
@@ -764,32 +764,13 @@ class PatchBundleRepository(
         if (!isEnabled) return
 
         // Trigger update so the new channel takes effect immediately.
-        // Additionally revert the preference so the toggle snaps back to its previous state.
-        val updateSucceeded = runCatching {
-            performRemoteUpdateWithResult(
-                force = true,
-                showToast = false,
-                allowUnsafeNetwork = false,
-                onPerBundleProgress = null,
-                predicate = { it.uid == uid }
-            )
-        }.isSuccess
-
-        if (!updateSucceeded) {
-            val reverted = prefs.bundlePrereleasesEnabled.get().toMutableSet()
-            if (usePrerelease) reverted.remove(uid.toString()) else reverted.add(uid.toString())
-            prefs.bundlePrereleasesEnabled.update(reverted)
-
-            dispatchAction("Revert prerelease ($uid)") { state ->
-                val src = state.sources[uid] ?: return@dispatchAction state
-                val reverted = when (src) {
-                    is APIPatchBundle -> src.copy(usePrerelease = !usePrerelease)
-                    is JsonPatchBundle -> src.copy(usePrerelease = !usePrerelease)
-                    else -> return@dispatchAction state
-                }
-                state.copy(sources = state.sources.put(uid, reverted))
-            }
-        }
+        startRemoteUpdateJob(
+            force = true,
+            showToast = false,
+            allowUnsafeNetwork = false,
+            onPerBundleProgress = null,
+            predicate = { it.uid == uid }
+        )
     }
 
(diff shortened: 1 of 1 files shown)


app/src/main/res/values/strings.xml
```diff
@@ -139,6 +141,11 @@ Are you sure you want to proceed?"</string>
     <string name="home_dialog_wrong_package_description">The selected APK is not for the application you intended to patch</string>
     <string name="home_dialog_expected_package">Expected package:</string>
     <string name="home_dialog_selected_package">Selected package:</string>
+    <string name="home_outdated_patches_dialog_title">Mobile data is active</string>
+    <string name="home_outdated_patches_dialog_message">Patch updates are disabled on mobile data. Your patches may be outdated</string>
+    <string name="home_outdated_patches_dialog_warning">Patching with outdated patches may produce an unstable result</string>
```
LisoUseInAIKyrios:
```suggestion
    <string name="home_outdated_patches_dialog_warning">Patching with outdated patches may result in a broken app</string>
```

app/src/main/java/app/morphe/manager/network/api/MorpheAPI.kt
```diff
@@ -257,30 +256,61 @@ class MorpheAPI(
...
+     * |----------------|-------------|---------|---------------|----------------|
+     * | 1.0.0          | off         | 1.0.0   | —             | nothing        |
+     * | 1.0.0          | off         | 1.1.0   | —             | 1.1.0          |
+     * | 1.0.0          | on          | 1.0.0   | 1.0.0-dev.1   | nothing        |
+     * | 1.0.0          | on          | 1.1.0   | 1.0.0-dev.1   | 1.1.0          |
+     * | 1.0.0-dev.1    | off         | 1.0.0   | —             | 1.0.0          |
+     * | 1.0.0-dev.1    | off         | 1.1.0   | —             | 1.1.0          |
+     * | 1.0.0-dev.1    | on          | 1.0.0   | 1.0.0-dev.1   | 1.0.0          |
+     * | 1.0.0-dev.1    | on          | 1.0.0   | 1.0.0-dev.2   | 1.0.0          |
+     * | 1.0.0-dev.1    | on          | —       | 1.0.0-dev.2   | 1.0.0-dev.2    |
```
MarcaDian: New logic

LisoUseInAIKyrios: The problem of dev not having the latest app-release.json is because of the two step release we are using, and semantic release doesn't back merge to dev.

But, jsdelvr now works and we can use it with manager to control when it gets the updated app-release.json

MarcaDian:
Not sure if this works now when the changes haven't been applied to the JSON on the backend yet. Or is it applied?

Edit: Ah, I see that the link to the dev branch still contains the dev version

LisoUseInAIKyrios:
Is this ready to merge?

MarcaDian:
Probably yes, it needs to be tested, especially after changes to FCM notifications.

LisoUseInAIKyrios:
Ok squash when you think it's ready. Can test in pre-release.

MarcaDian:
Note for the future:

Android often recreates activities when the application is minimized and restored, so the selected language is reset to the fallback locale. And when at this time a notification arrives about an available patch update or manager (in earlier PR, opening the manager by tapping on the notification was added), if the application is already running - then the update check logic should start, which in turn starts downloading the update.

Somewhere at this point it stops working. I think this is due to the activity being recreated from the background.

PR #248, opened 2026-02-25
fix: Remove duplicate string

BlackGold8282:


MarcaDian:
Thank you for your contribution!)

PR #282, opened 2026-03-04
fix: Root installation fails if module path does not exist

MarcaDian:
Fix https://github.com/MorpheApp/morphe-manager/issues/260#issuecomment-3993215390


MarcaDian:
Сlosed because it doesn't solve any of the problems.

LisoUseInAIKyrios:
This appears to be an issue that only affects Android emulators, and the app still works it just prompts every time.

PR #303, opened 2026-03-10
chore: Remove apkeditor as dependency

wchill:
We're not using apkeditor, so no reason to keep it in the repo.


MarcaDian:
Have you tested this? can we merge?

wchill:
I tested it locally when I opened the PR, but don't recall how thoroughly I tested it. will test it again before merging

PR #310, opened 2026-03-11
feat: Refine update badges using changelog scope matching

MarcaDian:
## Changes in this PR
Update badges now only appear when the bundle changelog contains changes scoped to the specific installed app, rather than on every bundle version bump.

## How it works
When a bundle update is detected, the app fetches the remote changelog and checks whether any newer entries contain changes for that specific app. Each changelog bullet is parsed for a conventional-changelog scope prefix (e.g. `**YouTube - Hide ads:**`), and matched against the app's display name via `startsWith`. If no matching scope is found, no badge is shown.

If the changelog is unavailable (network error, local bundle, or unknown app package), the badge is shown as a fallback to avoid silently missing updates.

### DIFF
--- a/app/src/main/java/app/morphe/manager/patcher/patch/BundleAppMetadata.kt
+++ b/app/src/main/java/app/morphe/manager/patcher/patch/BundleAppMetadata.kt
@@ -6,6 +6,7 @@
 package app.morphe.manager.patcher.patch
 
 import androidx.compose.ui.graphics.Color
+import androidx.compose.ui.graphics.toArgb
 import app.morphe.manager.util.KnownApps
 import app.morphe.patcher.patch.ApkFileType
 
@@ -100,12 +101,10 @@ data class BundleAppMetadata(
         // TODO: Remove once all active bundles ship Compatibility with appIconColor field.
         //  Transitional fallback for the period between Manager 1.3.0 release and
         //  patch bundles being updated to use the new Compatibility API.
-        private fun legacyAppIconColor(packageName: String): Int? = when (packageName) {
-            KnownApps.YOUTUBE       -> 0xFF0000  // YouTube red
-            KnownApps.YOUTUBE_MUSIC -> 0xFF0000  // YouTube Music red
-            KnownApps.REDDIT        -> 0xFF4500  // Reddit orange-red
-            KnownApps.X_TWITTER     -> 0x000000  // X black
-            else                    -> null
-        }
+        private fun legacyAppIconColor(packageName: String): Int? =
+            KnownApps.fromPackage(packageName)?.brandColor?.let { color ->
+                // appIconColor spec uses 0xRRGGBB - strip alpha from ARGB
+                color.toArgb() and 0x00FFFFFF
+            }
     }
 }
--- a/app/src/main/java/app/morphe/manager/ui/viewmodel/HomeViewModel.kt
+++ b/app/src/main/java/app/morphe/manager/ui/viewmodel/HomeViewModel.kt
@@ -489,9 +489,11 @@ class HomeViewModel(
         val updates = mutableMapOf<String, Boolean>()
 
         installedApps.forEach { app ->
+            // Get stored bundle versions for this app
             val storedVersions = installedAppRepository.getBundleVersionsForApp(app.currentPackageName)
             val appName = resolveChangelogName(app.originalPackageName)
 
+            // Check if any bundle used for this app has been updated
             val hasUpdate = storedVersions.any { (bundleUid, storedVersion) ->
                 val currentVersion = currentVersionByUid[bundleUid] ?: return@any false
                 if (!isNewerVersion(storedVersion, currentVersion)) return@any false
--- a/app/src/main/java/app/morphe/manager/util/Constants.kt
+++ b/app/src/main/java/app/morphe/manager/util/Constants.kt
@@ -56,18 +56,22 @@ object KnownApps {
      *
      * @param packageName The app's package name.
      * @param isPinnedByDefault Whether this app appears pinned on the home screen by default.
+     * @param brandColor App brand color used for the home screen button gradient start and
+     *   shimmer placeholder. Should match the appIconColor value shipped in the bundle's
+     *   Compatibility declaration. Null means fall back to [DEFAULT_COLORS].
      */
     data class Entry(
         val packageName: String,
         val isPinnedByDefault: Boolean = false,
+        val brandColor: Color? = null,
     )
 
     /** All known app entries in display order. */
     val all: List<Entry> = listOf(
-        Entry(YOUTUBE,       isPinnedByDefault = true),
-        Entry(YOUTUBE_MUSIC, isPinnedByDefault = true),
-        Entry(REDDIT,        isPinnedByDefault = true),
-        // Entry(X_TWITTER),  // Uncomment when release
+        Entry(REDDIT,        isPinnedByDefault = true, brandColor = Color(0xFFFF4500)),
+        Entry(YOUTUBE,       isPinnedByDefault = true, brandColor = Color(0xFFFF0033)),
+        Entry(YOUTUBE_MUSIC, isPinnedByDefault = true, brandColor = Color(0xFFFF0000)),
+        // Entry(X_TWITTER, brandColor = Color(0xFF000000)),  // Uncomment when release
     )
 
     // Fast lookup map - built once at startup.
@@ -78,11 +82,15 @@ object KnownApps {
 
     /**
      * Ordered list of shimmer placeholder gradient colors shown during cold-start loading.
-     * Uses [DEFAULT_COLORS] for every pinned slot — actual bundle colors will replace them
-     * once the bundle loads.
+     * Uses each app's [Entry.brandColor] as the gradient start — actual bundle colors will
+     * replace them once the bundle loads. Falls back to [DEFAULT_COLORS] if no brand color
+     * is declared.
      */
     val DEFAULT_SHIMMER_GRADIENTS: List<List<Color>> by lazy {
-        all.filter { it.isPinnedByDefault }.map { DEFAULT_COLORS }
+        all.filter { it.isPinnedByDefault }.map { entry ->
+            entry.brandColor?.let { color -> listOf(color, GRADIENT_MID, GRADIENT_END) }
+                ?: DEFAULT_COLORS
+        }
     }
 
     /**


MarcaDian:
This should be tested

optimatrix0:
I sincerely appreciate it! I’m very happy that you accepted the idea that both I and the user who suggested the changelog brought up.

If you want, you can send me the link here and I’ll test it without any problem. Or you can point me to the subsequent GitHub action so I can download it and give you feedback as updates come out.

I said I could help test it and I’ll gladly do that.

I also really liked the alternative solution you created. I was just thinking about that fallback you added: if there are any problems or failures with network updates, maybe another type of badge could appear, something like “waiting for update”. That way users won’t be completely confused.

This solution you created is also great because it brings together several alternative sources. Just like when you expanded the changelog to show multiple experimental changes, that was excellent. It really opens things up to multiple sources instead of relying only on a proprietary API from a single organization.

I’ll test it when you send it. I might take a little while because I need to find some time, so apologies in advance for that. But I already wanted to thank you because it’s turning out really amazing. Truly.

LisoUseInAIKyrios:
The existing fallback solution is needed if a third party repo has a weird changelog that doesn't follow semantic commit format, or it has some other formatting error.

It's better to fallback on "not sure, assume the updates patches include this app".

optimatrix0:
The organization behind ReVanced, which is somewhat the opposite side of this project, created an API to manage changelogs, changes, and everything else. Even so, it doesn’t come close to the kind of implementations you’re building here. Doing all of this without depending on an API, with everything processed locally and directly inside the Manager, and the Manager itself analyzing text to detect badges, is honestly incredible.

I also hope that someday it might be possible to support fixing the YouTube APKM from the APKMirror website, since that could be a really interesting addition. But that’s just something I hope for in the future — not something I’m trying to push or demand. I see it more as a natural evolution that could happen someday.

But seriously, for something that has only been developing for about two or three months, this is genuinely impressive.

I do have more ideas, but I’ll create proper feature suggestions later. For now, I just want to say thanks.

MarcaDian:
Artifact https://github.com/MorpheApp/morphe-manager/actions/runs/22956603189/artifacts/5871696991

optimatrix0:
I tested the artifact and installed it. Since I didn’t notice that you had sent the link earlier, I ended up grabbing it myself from the GitHub Actions.

I saw that you created a dedicated branch for this, which will actually be very useful for me to keep track of the progress.

I’ll need a bit of time to properly test it while new patch updates come out, just to make sure everything continues working correctly. So if you don’t mind a delay of maybe a day or two for feedback, I’d really appreciate it.

I also hadn’t thought about that use case with alternative patches having different logos. That was actually a very good point.

In the end, this is exactly what I was hoping for: through discussion we can reach a solution that works well for both sides — from the user perspective and from the developer perspective.

I’ll keep testing it, and once I have solid feedback I’ll let you know how it behaves. If I notice anything that might need adjustment, I’ll also report it.

optimatrix0:
I ran the test just now. It’s around a little after 5 AM here. From what I can see, everything is working well.

I updated to patch version v1.19.0-dev.1 (2026-03-12) and the behavior looks correct. Reddit does not show an update badge, which makes sense since there are no updates for it. Meanwhile, YouTube does show the update badge, because the changelog clearly lists several fixes and improvements for it.

So from my testing, the detection appears to be working as intended. As far as I can tell, this should be safe to merge into the next development version.

For context, I’m currently using a Starlink connection, so my Wi-Fi connection is stable. Only my mobile data connection tends to be slow when I need to rely on it.

optimatrix0:
I still think the idea of adding metadata in the patchlist file is relevant. The main reason is that it could act as a fallback mechanism.

As we all know, things can change over time. For example, something in the changelog parsing or fetch logic could break in the future if the semantics of entries for apps like YouTube or Reddit change. Because of that, relying 100% on one method may not always be ideal.

From my perspective, metadata could complement the current approach. The current system is working very well, which is great, but having metadata available as a fallback could increase long-term reliability and potentially reduce future support issues.

One of the developers here suggested the metadata approach as an alternative method, and I do agree with several aspects of that idea. The only difference in my view is that it might be even more useful if it worked as a secondary fallback rather than replacing the current system entirely.

If the team prefers to keep the current implementation as it is, that’s completely fine as well, especially since it’s already working well. This is simply a collaborative suggestion to potentially improve reliability in the future.

If something like this were also documented for the Manager, alternative patch sources could choose to adopt either method or both, which could benefit the wider community.

optimatrix0:
This PR may need one more change before it can be merged, based on my testing.

Before installing the artifact build, apps like Reddit were showing the update badge even though they should not have. This happened because the separation logic introduced by this PR was not yet active in the normal development build.

After I installed the artifact, the badge was still showing. What I did next was patch Reddit again. Once the patches were updated, the Manager correctly recognized that there was no update and the badge disappeared.

So the issue seems to be that when installing the artifact over an existing installation, the Manager does not immediately reanalyze the previously stored data. Ideally, it should refresh or reload the state after the update so that the badge reflects the correct status right away.

Because of this, it appears that a small refactor may be needed — possibly triggering a reload or re-evaluation of the stored patch/app state after the Manager updates.

That seems to be the only remaining issue I noticed during testing that may need to be addressed before merging this PR.

For example, when I updated again to version 1.0.13-dev.4, the badge appeared again because this PR had not yet been merged into that build.

Other than that, the feature works well. Test completed.

optimatrix0:
If this issue is resolved, then based on my testing this PR should be ready to be merged.

LisoUseInAIKyrios:
Clearing any previous "app update available" is not important because the logic will correct itself the next time the user patches.

The logic is also correct for apps patched after this change. Any migration code will be a one time thing and it's not worth the maintenance or future need to remove it

MarcaDian:
This seems to work fine for the default source, and theoretically works for third-party ones.

optimatrix0:
If everything is working as expected, then this PR can be merged into the next development update without any issues.

From my testing, the feature already appears to be working correctly. Earlier I suggested possibly reloading or re-evaluating the stored badge state after updating the Manager, but since that approach was not adopted and the fallback method was also not included, that’s completely fine. It seems the current implementation is already behaving correctly in practice.

At this point, I would just like to know if there is anything else within the scope of this PR that still needs testing. If there are specific scenarios or edge cases you would like verified, I’m happy to test them.

Otherwise, from my perspective as a tester, everything appears to be working well.

optimatrix0:
I think I understand the direction you prefer for the project now. Over the last few discussions I shared several ideas and feature suggestions because I genuinely wanted to help improve the Manager as much as possible.

From the responses so far, it seems the focus of the project is more on keeping things clean, simple, and easy to maintain, rather than adding too many new features that could increase complexity or support workload. That’s completely understandable.

I’m not criticizing the team at all. If anything, this helped me understand the project philosophy better. I probably got a bit too enthusiastic with my suggestions, and that’s on me. I just wanted to contribute and help where I could.

From now on I’ll try to focus more on smaller improvements and incremental ideas that better match the direction you prefer.

If possible, I would also really appreciate some guidance from the team about the kind of feedback or suggestions that are most helpful for you. Since these discussions are public, it might help other users as well to understand what type of ideas, improvements, or reports are most useful for the project.

Either way, I’m still happy to keep testing builds and sharing feedback when needed.

optimatrix0:
Since this PR is already touching the interface, I would like to suggest a small incremental improvement.

When using the pull-to-refresh gesture, besides checking for patch updates, the Manager could also check for updates for the other components in the menu. That was originally part of the idea when I discussed this topic before. At the time it seemed to be understood as something only related to patches, but the intention was broader.

Another small idea would be to make the text that appears during the pull-to-refresh change randomly. For example, when the user pulls down to refresh, the phrase at the top could rotate between different messages.

These phrases could also be expanded to include small tips about how to use the app. For example, tips like how to hide apps by long-pressing them, or other small usability hints.

I’m referring to the same type of messages that already appear when installing or opening the app, like “Which app do you want to patch?”, “Do you like ads? I don’t either.”, or “Saving apps from ads since now.” These kinds of phrases could rotate each time the user refreshes the screen.

Some tips could also appear occasionally, such as “Pull down like in a browser to check for updates.” Small hints like this could help users discover features naturally.

In my opinion, this could help users learn how the app works without needing a full onboarding screen. It’s a small incremental idea that could fit within this PR and improve usability without increasing complexity or support needs.

optimatrix0:
Another small thought about this idea: this could also become a fun little easter egg in the app.

The pull-to-refresh gesture is already something users interact with often, so having different messages appear there could make the experience feel a bit more alive. I can easily imagine the community on Reddit finding that kind of small detail really cool.

Instead of the text automatically cycling after a timer like some apps do, tying it to the refresh gesture lets the user control when they see the messages. That way they can actually read them instead of them disappearing too quickly.

Those messages could sometimes include small tips about the app, and sometimes just fun lines like the ones already present in the Manager. And over time you could even add occasional seasonal or special messages if you ever wanted to — things related to holidays or special days, for example.

It’s a tiny detail, but sometimes those little touches make an app feel more friendly and memorable. Just sharing the idea in case it fits the direction of the project.

MarcaDian:
At this time, I'm not sure if update notifications for third-party bundles are working correctly.

MarcaDian:
Displaying helpful hints on the main screen or during patching makes sense, and may be implemented in the future, but it is not currently the case for this PR.

It is very difficult to add different changes that are specifically different from each other. It is better to give changes in small portions, so it is better to test and debug if something goes wrong.

There are also a lot of requests from users, and few developers (there's only me right now), but we're all people, and we have personal lives, so we can't implement it as quickly as users would like.

optimatrix0:
Thank you very much for taking the time to respond and talk about my suggestions. I’m going to separate some of these ideas into different issues so things are easier to track, debug, and discuss. I think that will make it clearer for both the team and the community.

Now going back to the original topic about update notifications, I just want to clarify something. I’m not sure if you were referring to the update badges or the push notification alerts.

If you mean the push notification that appears for updates, as far as I understand that currently only works for the official package, since other sources haven’t implemented notification support. Unless the Manager starts actively checking repositories in a similar way to tools like Obtanium, indexing them and generating notifications, that probably won’t happen anytime soon.

But if you were referring to the update badges for third-party apps, based on my tests they appear to be working correctly.

One more thing I wanted to add to this discussion: even if some parts are not fully supported yet by third-party repositories, it may only be a matter of time before things align. Once the Manager implements the functionality, it’s quite possible that issues will start being opened in those external repositories as well so they can support it properly.

Open source communities tend to help each other in these situations. When maintainers understand the purpose and see users asking for it, they often implement the necessary changes without much trouble.

From my perspective, the idea already seems mature enough to move forward. If needed, I could also open issues in the relevant third-party repositories to point them to this discussion and explain the feature so they are aware of it.

MarcaDian:
Yes, I meant the badges in the buttons. Notifications for third-party bundles are currently not possible due to the current implementation, this is a process of interaction between our various repositories.

Initially, we developed the manager for our own needs at Morphe, and adding third-party bundles is currently in its basic implementation. For those who want a universal manager that covers all bundles, it is worth looking at other projects, such as URV.

It makes no sense to make a manager that will simply be a copy of another.

optimatrix0:
@MarcaDian

Thanks a lot for your words here and for explaining things the way you did, especially regarding what the project can realistically handle and your workload as the current main developer of the Manager, alongside our leader @LisoUseInAIKyrios.

We can keep the current requests I made as they are for now, but moving forward I’ll take your workload into consideration, especially knowing that you’re the only developer actively working on this repository. I honestly had no idea about that, so this helps me better understand the direction and limitations.

I also believe that third-party patches and sources tend to align over time. Clearly, without alternative sources, this project wouldn’t have the user base it has today. A good example is Google Photos, where many users were hesitant to migrate without proper support.

At this point, this request already seems perfectly fine according to your standards to be merged. Our leader @LisoUseInAIKyrios has already linked the issue to be closed once this PR is merged, so everything looks basically ready from here.

I do have Universal RV Manager installed, but honestly I prefer to support and follow the evolution of this project, which is why I’m focusing on using it instead of relying on other managers for third-party repositories.

Thanks again for the clarification.

optimatrix0:
@MarcaDian

If you’re up for it, could you share some guidance or basic documentation about how third-party sources should structure things so the Manager can correctly detect update badges?

After this PR is merged, if I notice that some third-party sources are not compatible (for example, not following the expected semantics), I’ll likely report it to them. So it would really help to understand what the “correct” implementation looks like from your side, so I can point them in the right direction.

Since everything on the Manager side already seems ready, I can help with that in my free time when possible. Of course, I might not always be able to do it immediately, but I’ll do my best to contribute where I can.

If you can provide even a simple explanation or reference, that would already be super helpful for opening issues in the appropriate third-party repositories.

Also, if you prefer, this could be added to the documentation in the future when you have time. That way everything would be well structured, and new developers creating sources would already know how to implement things correctly so the update badges work as expected. This could also help reduce support load on your side over time.

I understand if this might feel like extra work, but having clear documentation could go a long way in helping the community handle things more independently.

I also want to share a bit of my perspective. I’m just a 22-year-old from Brazil with visual impairment, using voice transcription to participate here. I’m still learning and doing my best to contribute within my limitations, but I genuinely want to help the project grow.

If there’s ever a way I could help more directly — for example, testing features related to usability or general behavior (even if not very visual) — I would be really happy to be part of that in any way possible.

Thanks again, I really appreciate your time and all the work you’re putting into the Manager 👍

MarcaDian:
Theoretically, this PR should also work for third-party bundles that use the Morphe patches repository template, if they make release and the changelog correctly.

The update badge will appear if the app name specified in the changelog matches the name of the installed or patched app.

Technically, how this works is already described in the first message.

PR #329, opened 2026-03-15
refactor: Use Morphe patcher 1.3.0

LisoUseInAIKyrios:
TODO:

- [x] Add basic usage of new Compatibility fields (app icon color, APK/APKM file type, etc)
- [x] Test

### DIFF
--- a/app/src/main/java/app/morphe/manager/ui/viewmodel/HomeViewModel.kt
+++ b/app/src/main/java/app/morphe/manager/ui/viewmodel/HomeViewModel.kt
@@ -51,6 +51,7 @@ import app.morphe.manager.util.PatchSelectionUtils.updateOption
 import app.morphe.manager.util.PatchSelectionUtils.validatePatchOptions
 import app.morphe.manager.util.PatchSelectionUtils.validatePatchSelection
 import app.morphe.patcher.patch.AppTarget
+import io.ktor.http.encodeURLPath
 import kotlinx.coroutines.*
 import kotlinx.coroutines.flow.*
 import kotlinx.datetime.TimeZone
@@ -1451,7 +1452,7 @@ class HomeViewModel(
 
         // Handle null pendingRecommendedVersion
         val escapedVersion = pendingRecommendedVersion?.let { encode(it.version, "UTF-8") } ?: "any"
-        val searchQuery = "$pendingPackageName:$escapedVersion:${Build.SUPPORTED_ABIS.first()}"
+        val searchQuery = "$pendingPackageName~$escapedVersion~${Build.SUPPORTED_ABIS.first()}".encodeURLPath()
         val searchUrl = "$MORPHE_API_URL/v2/web-search/$searchQuery"
         Log.d(tag, "Using search url: $searchUrl")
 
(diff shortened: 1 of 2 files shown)


optimatrix0:
@LisoUseInAIKyrios

Would it be possible to add a small piece of information in the “About” screen of the app showing something like:

“This version of the Manager uses Morphe Patcher X.X.X”

I think this could help with clarity and make the modular nature of the project more visible to users.

Since the Patcher version may change over time, this could also be updated automatically during the build process so each release reflects the correct version being used.

Any clearer information for users is always helpful. This could also allow more experienced users to compare versions themselves when patch files are updated and decide if they want to update or not.

Also, this is a bit unrelated, but I just wanted to say thanks for reducing the patch package size — that was really nice to see.

Let me know what you think. If it makes sense, I can open a separate issue for this as well.

LisoUseInAIKyrios:
The patcher version is always increasing and we won't revert back to an older version.

The information isn't important or useful to end users, and it may confuse them to why it's being shown.

optimatrix0:
@MarcaDian @LisoUseInAIKyrios

Man, thanks for implementing that suggestion about changing the phrases, seriously. It turned out really cool and it’s going to be a nice addition to the app. Maybe in the future, if you think it makes sense, this could even be expanded to show small usage tips too, but that’s up to you and your time.

About my “Welcome to Morphe” request, you can close it or mark it as not planned, no problem at all. I’ll keep using the app and only suggest things that actually make sense in real usage.

About showing the Morphe Patcher version, the idea is mainly to make it clear that the project is modular. It also helps with transparency. If any external organization ever questions the project, it becomes clearer that the Manager itself is not doing everything directly, but using a separate library.

A lot of apps do something similar, like showing modules in the app info, so it’s not unusual.

I know this may not be essential for most users, but at the same time, a lot of the information already shown today is also not essential and was still added because it helps with transparency.

I’m not trying to go against your vision or anything, just sharing the idea. And I also understand that constantly updating this kind of info in the UI might not be a priority.

Also, thanks for the translations, that was something that was really missing.

And honestly, part of all this is also about the “feel” of the app. Those phrases, the interface, the way everything is presented… it gives a kind of high-tech, almost sci-fi vibe, and that’s really cool. When I first installed the app, I felt that, and I think other users feel it too.

Users like nice interfaces. Users like information. Users like understanding what’s going on. That also increases trust in the project.

And just adding one more point: when we patch apps like YouTube, a new section appears inside the app, and in the “About” section it shows a lot of information — patch versions, whether it’s pre-release, official links, and even warnings if things are outdated. That shows this kind of information is not completely unnecessary.

Maybe adding something similar in the Manager’s “About” section, like the Morphe Patcher version or patch info, could help more advanced users understand what they’re using.

But at the same time, thinking about it again, it might also be a bit unnecessary for most users, just like the “About” section inside YouTube 😄

About what I said earlier regarding updates, just to clarify: I didn’t mean rollback or anything like that. I meant more about cases where users don’t update their patches, and having some way to visualize that.

Anyway, it’s just an idea, nothing mandatory. It’s really good to talk with you about these things, it even helps relieve some daily stress 😄

I’ll keep reporting issues when I find them, testing things, and suggesting small improvements when it makes sense. If you ever need someone to test something, feel free to call me.

That’s it for today. I hope I’m being at least a little useful around here 😄

Who knows, maybe tomorrow I’ll show up with another “problem” I created haha

optimatrix0:
I saw your contributions to this PR, @MarcaDian and @LisoUseInAIKyrios.

I couldn’t find any toggle to enable experimental versions. I’m not sure if it works differently, but I didn’t find anything on the app page, the settings screen, or the patch package page.

Is this feature not available yet in version 12.0.8? What do you think?

LisoUseInAIKyrios:
Experimental versions will come after Manager stable is released. Soon.

MarcaDian:
You can try to add test source locally with experimental toggle support.
Build-in source need to be disabled.

https://github.com/MorpheApp/morphe-patches/actions/runs/23397303326/artifacts/6044584990

PR #353, opened 2026-03-22
feat: Allow patching split APKs with a warning instead of blocking

MarcaDian:


LandonMoran:
I have had no issues with split APKs but I know they can introduce problems sometimes

PR #381, opened 2026-03-30
fix: Improve root mounting

MarcaDian:
Adapted from https://github.com/Jman-Github/Universal-ReVanced-Manager/blob/dev/app/src/main/assets/root/service.sh

### DIFF
--- a/app/src/main/java/app/morphe/manager/ui/screen/home/InstalledAppInfoDialog.kt
+++ b/app/src/main/java/app/morphe/manager/ui/screen/home/InstalledAppInfoDialog.kt
@@ -854,21 +854,25 @@ private fun ActionButtonsRow(
     actions: List<ActionItem>,
     isPrimary: Boolean
 ) {
-    Row(
-        horizontalArrangement = Arrangement.spacedBy(8.dp),
-        modifier = Modifier.fillMaxWidth()
-    ) {
-        actions.forEach { action ->
-            ActionButton(
-                text = action.text,
-                icon = action.icon,
-                onClick = action.onClick,
-                enabled = action.enabled,
-                isDestructive = action.isDestructive,
-                isPrimary = isPrimary && !action.isDestructive,
-                isLoading = action.isLoading,
-                modifier = Modifier.weight(1f)
-            )
+    Column(verticalArrangement = Arrangement.spacedBy(8.dp)) {
+        actions.chunked(2).forEach { rowActions ->
+            Row(
+                horizontalArrangement = Arrangement.spacedBy(8.dp),
+                modifier = Modifier.fillMaxWidth()
+            ) {
+                rowActions.forEach { action ->
+                    ActionButton(
+                        text = action.text,
+                        icon = action.icon,
+                        onClick = action.onClick,
+                        enabled = action.enabled,
+                        isDestructive = action.isDestructive,
+                        isPrimary = isPrimary && !action.isDestructive,
+                        isLoading = action.isLoading,
+                        modifier = Modifier.weight(1f)
+                    )
+                }
+            }
         }
     }
 }


MarcaDian:
I'm doing this practically blindly. So I'm not sure if it works or if it's bug-free.

MarcaDian:
Artifact to test
https://github.com/MorpheApp/morphe-manager/actions/runs/23744968448/artifacts/6178099675

Jman-Github:
> I'm doing this practically blindly. So I'm not sure if it works or if it's bug-free.

As far as I know users who mount all the time don't report any issues. I perfected it through constant user feedback, so it should be pretty stable.

MarcaDian:
@Rikysonic Can you try this artifact?

Rikysonic:
Just to clarify: I currently have no issues updating patches on top of already-patched app if no version upgrade of the original app is involved - the new patched apk applies just fine, even after rebooting.
The issue with unmount request is specific on how Android normally handles the app upgrades: Android creates a new folder in /data/app with the new app, then deletes the old folder. The problem is that the patched apk is mounted directly on top of the original one in the old folder, and basically locks it in read-only mode, so Android isn't able to properly cleanup the old version, leaving it there. Upon rebooting, for some reason the old folder in /data/app is prioritized and crashes appear unless trying to upgrade again (dunno how it fixes then, but well it does).
Unmounting the patched apk before upgrading ensures that the old folders are correctly cleaned up when installing the new original update, so it's not something to be fixed during the patching process, but before even applying the new original app update, and I guess only manual unmounting could help here, or somehow listen for the Android app update trigger and unmount just before it.

MarcaDian:
This PR adds an `Unmount` button, it would be great if you could check if it works properly.

Rikysonic:
Oh wow, I'm sorry, I thought this PR was related to root mount fixes only, I didn't notice the unmount feature was added 😅
Well, the unmount button works as expected, thx for you work!
One stupid thing I noticed: in log.txt "force-stop failed (ignored)" is always printed even if force-stop happened - that command is probably returning false every time.
Speaking of force close, I think Morphe should also force close the app during the patching process, right after mounting the new patched apk (this is related to the original Morphe code, not this PR; I wanted to point it out because I often have to manually kill the newly patched app, otherwise it starts stock).

MarcaDian:
Artifact 
https://github.com/MorpheApp/morphe-manager/actions/runs/23838043680/artifacts/6215916475

Rikysonic:
Force stop after patching works great.
Force stop in service.sh doesn't work anymore, it prints:
`force-stop failed for com.google.android.youtube (exit: 2)`
in log file and app is actually not force closed - it's like it's not able to pass the `$package_name` variable in the if clause, but I'm speculating.
Also, could you add force close in unmount button too?

MarcaDian:
Artifact 
https://github.com/MorpheApp/morphe-manager/actions/runs/23861283945/artifacts/6226060220

MarcaDian:
If this does not work, please try manually checking the command parameter:
https://github.com/MorpheApp/morphe-manager/blob/d795cb9e585f8010f11b59489ce6bc7d5ebd6f12/app/src/main/assets/root/service.sh#L108

Although it is strange that such an identical command works in buttons.

Rikysonic:
Ok so, regarding the force-stop in `service.sh`: after lots of tests, it turns out that... it never worked, and it still doesn't. There's an error inside log that says `cmd: Failure calling service activity: Failed transaction (2147483646)`, and it also appears in stock Morphe - I thought it was related to something else in the script, since it's so generic, but moving the `am force-stop` code to different parts of the script, and adding random echos before and after it makes it clear that the error is specific to the force-stop code.
So I tried removing it altogether and... everything works just fine: after reboot, the app is correctly mounted and ready to use. I guess you can safely remove it.
Funny thing: if you place the force-stop code in a random sh file inside /data/adb and then run it with `su -mm -c "'/data/adb/random-script.sh'"`, for some reason it kills the app just fine - I really have no clue why... maybe you cannot run force-stop commands inside service scripts in modules? I don't really know.
Regarding the force-stop after unmount in Morphe app: it works, now the app is basically always force-closed after every mount/unmount event.
Thank you so much for the hard work!

LisoUseInAIKyrios:
Magisk work no issues as far as I can tell. Restarting device and patched app is still patched.

MarcaDian:
If this PR doesn't break anything then it's worth a merge because it adds an `Unmount` button and force closes the app when mounting. 

However, does it display 2 buttons Remount /Mount / Unmount in a row? can someone post a screenshot app info dialog?

Rikysonic:
App mounted:
Screenshot_2026-04-02-00-05-38-911_app morphe manager
App unmounted:
Screenshot_2026-04-02-00-05-47-890_app morphe manager

NMBlanco:
I'm having issues with this new update to mount installing.
I tried patching Reddit as usual, patching goes without errors, but in the end I don't get asked anymore how I would like to install the patched app (root install, normal install). It says I have to uninstall the currently installed apk, even though I chose root mount installer in settings.

Using KernelSU on OxygenOS Android 15

MarcaDian:
What you write about was not edited in this PR.

This PR only changed the mount script and restored the Unmount button.

Try to clear manager's cache or clear app data.

PR #393, opened 2026-04-02
fix: Use card background as color preview, add transparency checkerboard

MarcaDian:
Replace color preview dots with full-card color backgrounds. Cards now display the actual patch color as their background with adaptive text contrast based on effective luminance. Semi-transparent colors show a checkerboard underlay. Material You options use a subtle gradient.
___

| Dark | Light |
|--------|--------|
|  |  |


ILoveOpenSourceApplications:
Can we have screenshots from manager in light mode as well?

MarcaDian:
<details>

PR #409, opened 2026-04-08
fix: Read stored language directly from `SharedPreferences`

MarcaDian:


MarcaDian:
Closed because it does not solve the problem https://github.com/MorpheApp/morphe-manager/issues/408

PR #443, opened 2026-04-17
feat: Add toggle to disable home screen patching phrases

jaival-11:
### Description
This PR adds a toggle in the Appearance settings allowing users to show or hide the witty "patching phrases" on the Home Screen header. This is helpful for users who prefer a cleaner, more minimal UI. This feature was built to resolve a community feature request from the Morphe subreddit: [Fair request ig](https://www.reddit.com/r/MorpheApp/comments/1snzfl5/fair_request_ig/)

### Changes Made
* Added a `Switch` in `AppearanceTabContent.kt` under a new "Home Screen Header" section.
* Wrapped the `greetingMessage` initialization and refresh logic in `HomeScreen.kt` with a conditional check.
* **Note on Implementation:** To keep this PR isolated and avoid accidentally breaking the core `PreferencesManager` domain logic, I utilized standard Android `SharedPreferences` directly within the Compose UI state. 

### Releases
I made a release of the apk with this code (signed with debug key) [here](https://github.com/jaival-11/morphe-manager/releases/tag/v1.0-custom-toggle)  

### Transparency Note
I am a student stepping into open-source Android development. To ensure I didn't break the build environment, I used an AI assistant to help format the Kotlin/Compose syntax and execute the patch via terminal. I have locally built and fully tested the APK to verify it works exactly as intended. I have also used AI to write this PR. I am new to github, and this everything. I am still learning, please forgive me. I apologise for it. If I made any mistake, please correct me. Thank you 🙇🏻‍♂️

If the maintainers prefer this toggle to be integrated deeply into the `PreferencesManager` data store instead of localized `SharedPreferences`, please let me know and point me in the right direction—I am here to learn and happy to make adjustments!


jaival-11:
@MarcaDian I tested the built apk with this code on my secondary low end device (as I didn't want to unistall morphe on my main device). So I installed morphe, and check if toogle works as intended, and it persists even if you close and open the app again. I didn't tried to patch any app there.

app/src/main/java/app/morphe/manager/ui/screen/settings/AppearanceTabContent.kt
```diff
@@ -76,6 +76,34 @@ fun AppearanceTabContent(
             onLanguageClick = { showTranslationInfoDialog.value = true }
         )
 
+        val sharedPrefs = context.getSharedPreferences("morphe_settings", android.content.Context.MODE_PRIVATE)
+        var showPatchingPhrases by remember { mutableStateOf(sharedPrefs.getBoolean("show_patching_phrases", true)) }
+
+        SectionTitle(
+            text = "Home Screen Header",
```
MarcaDian: This should be in `res/strings.xml`

app/src/main/java/app/morphe/manager/ui/screen/settings/AppearanceTabContent.kt
```diff
@@ -76,6 +76,34 @@ fun AppearanceTabContent(
...
+        )
+
+        RichSettingsItem(
+            onClick = {
+                val newValue = !showPatchingPhrases
+                showPatchingPhrases = newValue
+                sharedPrefs.edit().putBoolean("show_patching_phrases", newValue).apply()
+            },
+            showBorder = true,
+            title = "Patching Phrases",
```
MarcaDian: This and subtitle also should be in `res/strings.xml`

app/src/main/java/app/morphe/manager/ui/screen/HomeScreen.kt
```diff
@@ -57,13 +57,14 @@ fun HomeScreen(
     val isRefreshing by homeViewModel.isRefreshing.collectAsStateWithLifecycle()
 
     // Get greeting message
-    var greetingMessage by remember { mutableStateOf(context.getString(HomeAndPatcherMessages.getHomeMessage(context))) }
+    val sharedPrefs = context.getSharedPreferences("morphe_settings", android.content.Context.MODE_PRIVATE)
```
MarcaDian: This can be passed through the `HomeViewModel`

app/src/main/java/app/morphe/manager/ui/screen/HomeScreen.kt
```diff
@@ -57,13 +57,14 @@ fun HomeScreen(
...
+    var greetingMessage by remember { mutableStateOf(if (sharedPrefs.getBoolean("show_patching_phrases", true)) context.getString(HomeAndPatcherMessages.getHomeMessage(context)) else "") }
 
     // Handle refresh with haptic feedback
     val onRefresh: () -> Unit = {
         view.performHapticFeedback(HapticFeedbackConstants.LONG_PRESS)
         HomeAndPatcherMessages.resetHomeMessage()
-        greetingMessage = context.getString(HomeAndPatcherMessages.getHomeMessage(context))
+        greetingMessage = if (sharedPrefs.getBoolean("show_patching_phrases", true)) context.getString(HomeAndPatcherMessages.getHomeMessage(context)) else ""
```
MarcaDian: This can be passed through the `HomeViewModel`

MarcaDian (dismissed):

jaival-11:
@MarcaDian  Thank you for the review and guidance!

​I have made the requested changes:
1. Moved the hardcoded UI text to res/values/strings.xml and referenced them via stringResource.
2. Extracted the SharedPreferences logic out of HomeScreen.kt and passed the boolean state through HomeViewModel.

​Note: I haven't built apk and tested myself. Please tell if all working, or any changes to make

MarcaDian:
There is only one minor "problem" left. The greetings panel was also a placeholder for the update snackbars.
Now they will overlap the main buttons

jaival-11:
@MarcaDian ​I just pushed a quick fix that replaces the empty string "" with a blank space " ". This keeps the Compose Text node at its exact standard single-line height, preserving the invisible bumper spacing so the snackbars don't overlap the layout..

Please tell if this is what is expected, or another change is to make

MarcaDian:
Still, the implementation of the preference is incorrect, you need to look at similar code in homeviewmodels, also why specify `" "` if we just need to hide this greetings block. Later I will see how to implement it correctly.

jaival-11:
@MarcaDian I made some changes as you said. Please tell if this time Implementation is correct or not

MarcaDian:
You have duplicate lines of code.

jaival-11:
So sorry for that. I was editing via termux and pasted same code twice. I think I made necessary changes. Please tell any issues in the code. Sorry once again for that

DonDae001:
@jaival-11 Would it be possible to put the Morphe logo on top as a substitute for the missing headers?

<details>
<summary> Currently, it looks empty and wasted space (see here). </summary>
Morphe&#39;s Look
</details>

Here's a mockup of what it could look like (apologies if the logo isn't 100% in the middle):

| M-only logo | Full name logo |
| --- | --- |
| Morphe Logo Header Mockup 01 | Morphe Logo Header Mockup 02 |

Of course, this could be toggled on and off, and changed to either of the logos. Thanks!

MarcaDian:
> Currently, it looks empty and wasted space (see here).

In my last changes, this space is no longer there.

MarcaDian:
@LisoUseInAIKyrios Why we can't build the artifact in this PR?

LisoUseInAIKyrios:
Actually, we can't run build pull request because there is no way to select a remote branch. The action would need to run on the other repo (and be setup with signing keys, etc).

jaival-11:
> > Currently, it looks empty and wasted space (see here).
> 
> In my last changes, this space is no longer there.

So shall I think of implementing the logo in header, as suggested?. If so, what are the changes made so I can understand them before trying to implement this.

jaival-11:
@DonDae001 Giving choice is always a good thing. I will think of implementing it. Thank you for your suggestion

MarcaDian:
I don't think this is necessary. The user meant to use a reserved space, but now it's gone. This could be a separate request.

jaival-11:
> I don't think this is necessary. The user meant to use a reserved space, but now they don't have one. This could be a separate request.

Sure, it makes sense. I may give it a thought in future.

So now without that space, how does it look? Is it like sufficient top margin so it looks normal (I hope you get what I am trying to say)

MarcaDian:
Now the buttons start at the very top. As I said, the snackbar overlaps the app buttons if third-party bundles are enabled.

LisoUseInAIKyrios:
> I don't think this is necessary. The user meant to use a reserved space, but now it's gone. This could be a separate request.

I don't think the app name as a header is needed either. The user knows what app they are using, and it's less desirable to add another setting to change this.

DonDae001:
> I don't think the app name as a header is needed either. The user knows what app they are using, and it's less desirable to add another setting to change this.

Well, YouTube has a header that says "YouTube" right after opening the app, similar to my concept.
Screenshot_20260419_174739_YouTube Morphe

MarcaDian:
> Well, YouTube has a header

This is the topbar, and besides the logo, it also contains search and settings buttons. If we add the Morphe logo, then the main problem of this PR is lost - saving space (or I misunderstood and users don't want to see funny greetings from the manager 😅)

MarcaDian:
@LisoUseInAIKyrios Can we merge or will we implement the proposal?
I want to create a new dev release now and am waiting for this change

DonDae001:
I feel like this could be an feature request issue instead. Should I make one (after dev release) so there could be a dev release now?

DonDae001:
>then the main problem of this PR is lost - saving space

I'm pretty sure some people just want to remove the text simply because it's 'force-fed AI slop'. From the Reddit post that @jaival-11 mentioned, the OP of the post said: "Just have a request that we abandon this **shitty slop generating headers.** "

LisoUseInAIKyrios:
Those headers were created by us without AI.

jaival-11:
@DonDae001, I can confirm they are just predefined strings, not ai. Your Implementation is not bad, but imo it needs some refinement. Like it may feel weird to just have header. May be besides header, the search button can be shifted there. So, there needs to be some time given for ideation on this. Suggestions are always welcome

@MarcaDian How does this affect landscape orientation, especially on tablets

jaival-11:
@MarcaDian, tested the pre-release. I think the top margin should be a little more. Rn, it is like at very top

MarcaDian:
<details>



</details>

This is standard lazy column padding 16.dip from all edges

jaival-11:
We can set custom padding, so bottom, left and right is 16 and top is more.

MarcaDian:
I can do it on dev, but what's the value?

jaival-11:
@MarcaDian  How about 24, or 32? I think 48 will be too much

PR #452, opened 2026-04-19
Clarify APK file location requirement

logm1lo:
Added instruction to move the APK file to the root directory after downloading for proper functionality.


MarcaDian:
This is a solution only for some vendors, such as Samsung or Xiaomi.

LisoUseInAIKyrios:
I have Samsung and I don't need to move the original APK file anywhere.

logm1lo:
so maybe it need a rephrase like: "If you can't select the apk file, move the apk file to the root directory not in the Download folder after downloaded"

logm1lo:
so im going to close this pr and then open another one

PR #457, opened 2026-04-20
feat: Improve patch visibility in bundle and app patch dialogs

MarcaDian:
Bundle names are displayed as styled pill badges.
- In `AppPatchesDialog` (multi-bundle): patch cards are tinted per bundle via deterministic HSL color, making it easy to visually distinguish which bundle a patch belongs to.
- In `BundlePatchesDialog`: patch cards are tinted using the app's `appIconColor` declared in the bundle, falling back to the default surface color when not available.
___

### DIFF
--- a/app/src/main/java/app/morphe/manager/ui/screen/HomeScreen.kt
+++ b/app/src/main/java/app/morphe/manager/ui/screen/HomeScreen.kt
@@ -51,10 +51,6 @@ fun HomeScreen(
     val view = LocalView.current
 
     // Dialog states
-    val showInstalledAppDialog = remember { mutableStateOf<String?>(null) }
-    // Increments on every open - ensures key(packageName, openToken) forces fresh
-    // composition even when the same package is opened twice in a row
-    val installedAppDialogToken = remember { mutableIntStateOf(0) }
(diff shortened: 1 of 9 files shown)


MarcaDian:
@LisoUseInAIKyrios What do you think about this PR?

LisoUseInAIKyrios:
Looks good to me. I think the card background color could be more subtle (a barely noticeable color change to the app icon), but this looks good too.

MarcaDian:
If increase the alpha even more, the cards becomes difficult to see in a dark theme.

MarcaDian:
Or, conversely, reduce:

<details>




</details>

LisoUseInAIKyrios:
I meant, keep the alpha the same, but can the color of the card be shifted towards the background color or some other color?

basically, use the app color as a slight tint to a different color of the card (same base color for all app cards)

ghost:
If you slide app on the homepage to view patches, they are still in the old colored background:

Screenshot_20260424_110546_Morphe

MarcaDian:
The color is not displayed if only the standard Morphe source is present. This makes no sense.

PR #479, opened 2026-04-25
chore: clarify APK file picker tip in README

logm1lo:
Clarified instructions for providing the APK file by adding a note about moving the APK to the root directory if it cannot be selected.


MarcaDian:
Thanks!

PR #483, opened 2026-04-25
feat: Open `.mpp` patch sources directly from file manager

### LINKED ISSUE #482: feat: Support direct import of .mpp files from file explorer
### Feature description

Hello Team,
I am a patch developer and usually build patches locally. Currently, the workflow is quite lengthy—I need to build the patch, manually add the patch files, open the app, and then import the files through the app interface.
I would like to request support for directly importing .mpp files from the file explorer into the Morphe Manager app. This would simplify the workflow and make the process faster and more convenient for developers

### Motivation

The current process of importing patch files is time-consuming and requires multiple manual steps. Directly importing .mpp files from the file explorer would simplify this workflow.
This feature would save time, improve developer productivity, and make patch testing faster and more efficient.

### Acknowledgements

MarcaDian:
Implements native file association for `.mpp` patch bundle files.
Tapping a `.mpp` file in any file manager now opens Morphe with a confirmation dialog to import it as a local patch source.
___

### DIFF
--- a/app/src/main/AndroidManifest.xml
+++ b/app/src/main/AndroidManifest.xml
@@ -84,16 +84,13 @@
                 <category android:name="android.intent.category.DEFAULT" />
             </intent-filter>
 
-            <!-- Open .mpp patch bundle files from file managers  -->
-            <intent-filter>
-                <action android:name="android.intent.action.VIEW" />
-                <category android:name="android.intent.category.DEFAULT" />
-                <data android:scheme="file" android:mimeType="*/*" />
-            </intent-filter>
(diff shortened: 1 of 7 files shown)


LandonMoran:
This is also good because if the manager ever has trouble grabbing patches or you want to downgrade your patches then this can help

MarcaDian:
You could always add a local bundle through the manager interface. This change makes it a little more convenient.

LandonMoran:
I'm aware but honestly after you download something the first thing you're going to do is open your file manager and like open the thing so like it just makes it way more intuitive

PR #563, opened 2026-05-18
fix: Skip stock app reinstall when same version is already installed for mount

MarcaDian:


kitadai31:
You may also want to fix this issue:

https://github.com/ReVanced/revanced-manager/issues/3288

`pm uninstall -k --user X` should not be used here.

MarcaDian:
Thanks, I'm not sure if I'm making these changes correctly, as I haven't dealt with root in over 10 years.

kitadai31:
Wait, please nvm

I had completely forgotten about the existence of the "-d" flag.
And apparently if you use the -d option with su permission, you can downgrade the app without uninstalling it.
Therefore, the situation like the "tracing" I mentioned in my issue in ReVanced repo will not occur.
(I've forgot to run those commands with superuser at that time.)
I'm sorry.

So, the above issue was half incorrect.
While "pm uninstall -k --user X" is indeed unnecessary, there was no need to perform the uninstallation beforehand in the first place.
You can run "pm install -r -d" directly without having to run "pm uninstall -k" beforehand.
No app data will be lost.
(Tested on my Pixel 3, LineageOS 22, CorePatch is not enabled)

I'll also correct my issue in ReVanced repo ASAP.

However, that raises another problem.
When I had the latest YouTube version from Play Store installed then installed an older YouTube version like 20.40.45 over it with `-d` option, YouTube crashed due to a downgrade.
Is it really desirable to retain the data with `-k` option?
Perhaps it would be better to ask user to allow uninstall the newer app through the Manager UI just like non-root install, rather than downgrading the app with retaining the data silently.

MarcaDian:
@kitadai31 Have you tested this PR? No user feedback. And I don't know if this change is ready for release.

kitadai31:
OK so, yes, I built this PR and tested

Mounting failed if the stock app is not installed or if a newer version is already installed.

### • If the stock app is not installed

Screenshot_20260522-142318_Morphe

### • if a newer version is already installed

Screenshot_20260522-142143_Morphe

("Mount" button does nothing here)

So the downgrading or installing stock apps isn't working

However, just to be sure, I also tried the current latest stable Manager (v1.17.1), and the result was the same. Therefore, this seems to be a separate issue unrelated to this PR.

In the situation where the same version is already installed, which is the target of this pull request, mounting worked successfully.

MarcaDian:
Closed for now, explanation []()https://github.com/MorpheApp/morphe-manager/issues/562#issuecomment-4519931455

PR #567, opened 2026-05-19
feat: add settings screen back button

### LINKED ISSUE #396: feat: Add a visible back button in the settings screen
### Feature description

Description:
Implement a visible back button in the settings screen, positioned at the top (e.g., top-right corner), to allow users to navigate back easily without relying on system gestures.

### Motivation

Motivation:
Some users—especially those using screen readers—experience difficulty navigating back using gestures or system buttons. In some cases, users may not understand the gesture or may have disabled system navigation buttons entirely.

Since accessibility is an important aspect of the app, adding a clear and visible back button would significantly improve usability. Additionally, the YouTube settings interface already includes a similar back button, so implementing this would also improve consistency across the app.

This feature would be relatively simple to implement and would provide a meaningful accessibility improvement for many users.

### Acknowledgements

ZaidNAlAsali:
## Summary

Adds a visible back button to the settings screen.

This gives users a clear way to return from settings without relying only on system gestures or navigation buttons, improving accessibility and discoverability.

Closes #396.

## Testing

* Ran `git diff --check`
* Confirmed the only `SettingsScreen` call site is updated
* Verified there is no existing reusable generic back/navigation string
* Verified `SettingsScreen` did not already have a top app bar/header
* Tried `.\gradlew.bat :app:compileDebugKotlin`, but local Android SDK configuration is missing:
  `SDK location not found. Define a valid SDK location with an ANDROID_HOME environment variable or local.properties.`

### DIFF
--- a/app/src/main/java/app/morphe/manager/MainActivity.kt
+++ b/app/src/main/java/app/morphe/manager/MainActivity.kt
@@ -304,7 +304,10 @@ private fun MorpheManager(vm: MainViewModel) {
             }
 
             composable<Settings> {
-                SettingsScreen(homeViewModel = homeViewModel)
+                SettingsScreen(
+                    homeViewModel = homeViewModel,
+                    onBackClick = { navController.popBackStack() }
+                )
             }
         }
     }
--- a/app/src/main/java/app/morphe/manager/ui/screen/SettingsScreen.kt
+++ b/app/src/main/java/app/morphe/manager/ui/screen/SettingsScreen.kt
@@ -15,10 +15,12 @@ import androidx.compose.foundation.pager.HorizontalPager
 import androidx.compose.foundation.pager.rememberPagerState
 import androidx.compose.foundation.shape.RoundedCornerShape
 import androidx.compose.material.icons.Icons
+import androidx.compose.material.icons.automirrored.outlined.ArrowBack
 import androidx.compose.material.icons.outlined.Palette
 import androidx.compose.material.icons.outlined.Settings
 import androidx.compose.material.icons.outlined.Tune
 import androidx.compose.material3.Icon
+import androidx.compose.material3.IconButton
 import androidx.compose.material3.MaterialTheme
 import androidx.compose.material3.Surface
 import androidx.compose.material3.Text
@@ -69,6 +71,7 @@ private enum class SettingsTab(
 @Composable
 fun SettingsScreen(
     homeViewModel: HomeViewModel,
+    onBackClick: () -> Unit,
     themeViewModel: ThemeSettingsViewModel = koinViewModel(),
     importExportViewModel: ImportExportViewModel = koinViewModel(),
     patchOptionsViewModel: PatchOptionsViewModel = koinViewModel(),
@@ -164,6 +167,8 @@ fun SettingsScreen(
             .fillMaxSize()
             .statusBarsPadding()
     ) {
+        SettingsHeader(onBackClick = onBackClick)
+
         // Content area
         HorizontalPager(
             state = pagerState,
@@ -218,6 +223,40 @@ fun SettingsScreen(
                     pagerState.animateScrollToPage(tab.ordinal)
                 }
             }
+        )
+    }
+}
+
+@Composable
+private fun SettingsHeader(
+    onBackClick: () -> Unit
+) {
+    Row(
+        modifier = Modifier
+            .fillMaxWidth()
+            .height(64.dp)
+            .padding(horizontal = 12.dp),
+        verticalAlignment = Alignment.CenterVertically
+    ) {
+        IconButton(
+            onClick = onBackClick,
+            modifier = Modifier.size(48.dp)
+        ) {
+            Icon(
+                imageVector = Icons.AutoMirrored.Outlined.ArrowBack,
+                contentDescription = stringResource(R.string.back)
+            )
+        }
+
+        Spacer(modifier = Modifier.width(8.dp))
+
+        Text(
+            text = stringResource(R.string.settings),
+            style = MaterialTheme.typography.titleLarge,
+            fontWeight = FontWeight.SemiBold,
+            color = MaterialTheme.colorScheme.onBackground,
+            maxLines = 1,
+            overflow = TextOverflow.Ellipsis
         )
     }
 }
--- a/app/src/main/res/values/strings.xml
+++ b/app/src/main/res/values/strings.xml
@@ -718,6 +718,7 @@ The image dimensions must be as follows:
     <string name="patches">Patches</string>
     <string name="home">Home</string>
     <string name="settings">Settings</string>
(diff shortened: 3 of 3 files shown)


LandonMoran:
Thank you so much, I was going to open an issue on this but I guess I don't have to being blind and all that, I physically don't know how to use the back gesture so I back button for settings is super nice though. To be fair, it wasn't a huge issue because I didn't enter the settings too often

MarcaDian:
This PR seems to be generated by AI. Currently, some redesign has been done in the dev version, but we don't have a concept for adding such a button to the current settings yet. The standard toolbar (added as I understood in this PR) is not suitable in terms of design. We are open to suggestions, preferably with a visual preview.

ZaidNAlAsali:
Thanks for the feedback. That makes sense.
I’ll revisit the approach and make sure it fits the current dev redesign instead of adding a generic toolbar/header.

________________________________

MarcaDian:
I tried adding the back button to the bottom tab bar, so that it wasn't in a separate header, I don't know how correct that is. This button is read first in Talkback as if it were at the top, so there shouldn't be any problems for blind users with screen readers.

<details>




</details>

LisoUseInAIKyrios:
A back button on the buttom nav bar is unusual, and I agree I don't think it would look good with a top nav bar either.

Maybe we will figure out a different solution that isn't an explicit back button at all.

MarcaDian:
As a separate solution for screen readers, we could add an invisible back button. But some sighted users might want a visible button, so this is a controversial solution.

LandonMoran:
Let you guys know that when a back button in the YouTube settings, like the custom settings menu you guys have, there is a back button. I can tell it's the back button, but not because it's labeled, specifically because it's the only unlabeled button. So I literally have to distinguish the back button by the fact that it's unlabeled, same with the refresh video button in the video player. That is basically the only issue I've seen. I'm still having the issue where occasionally when entering comments the application will crash. Although I still think that's an Android 17 bug, but it seems to be less frequent now so I don't know what's up with that. And earlier today I had a bug where literally when I turned up the volume it crashed my application

LandonMoran:
What I mentioned also has to do also occurs with the save to watch later button. I just forgot to mention that cuz I don't use that button

MarcaDian:
How does this apply to the Morphe Manager?

LandonMoran:
It doesn't but it's just something I'm pointing out because it has the exact same issue and I don't want the buttons here to be issue that the button's there have of no labels

ZaidNAlAsali:
Thanks for looking into this and for trying another placement.
I agree that both a separate top toolbar and a back button inside the bottom bar may not be the right fit visually. I opened this mainly because the accessibility/use-case seemed valid, but I’m happy to rework the PR around whatever direction fits the current dev redesign best, or close it if you prefer to handle the navigation pattern separately.
If helpful, I can also turn this into a smaller accessibility-focused change instead of a visible UI change.

________________________________

optimatrix0:
As an accessibility user, I think I can provide some feedback on this topic.

For screen reader users, an invisible "Back" button at the top of the page would already be an excellent solution. The screen reader would announce it, and activating it would return the user to the application's main screen. Personally, I would be completely satisfied with that approach. Many applications already use similar accessibility-only elements, and they work very well.

The challenge is providing a similarly intuitive experience for sighted users who may also benefit from having a convenient way to navigate back without adding visible UI elements.

One possible idea could be implementing a gesture. For example, when focused on the "System" button on the home screen, swiping upward could trigger a back navigation action. This would preserve the clean interface while still offering an accessible and discoverable navigation method.

Of course, this is just a suggestion and not something I'm strongly attached to. From my perspective as a screen reader user, the accessibility button alone would already solve the problem. I'm simply trying to think of an approach that could work well for both accessibility users and sighted users while keeping the interface minimal.

MarcaDian:
Added in https://github.com/MorpheApp/morphe-manager/commit/9753c49624859dbe81ba1747693ecb4edb9341b8

LandonMoran:
Developer release the back button is not present in the settings UI I just installed it

MarcaDian:
It is added invisible as suggested in https://github.com/MorpheApp/morphe-manager/pull/567#issuecomment-4741468633 (need test).

And it is present as a full button in tablet mode.

LandonMoran:
I am on the latest developer beta using TalkBack and it's not there. If I don't went to settings I get stuck in settings because there is no back button

MarcaDian:
Hmm, it should be. I need to look again.

I don't know how to check, how should it be? I go into the settings and it should be the first one to be announced? (this is how it was planned)

optimatrix0:
I tested the recently released Dev 4 version and noticed that the accessibility improvements I previously suggested were implemented. The screen reader actions and gesture support are working very well, and everything feels much more accessible now. Thank you for that.

However, regarding the issue mentioned by the previous user, I can confirm that the Back button still does not appear to be accessible. Using TalkBack 17.1, the button is not announced or exposed to the screen reader, so it seems to be ignored entirely from an accessibility perspective.

Other than that, the improvements are working great. I just wanted to report this remaining issue in case it helps.

Thanks!

LandonMoran:
It should be the first one to be announced. Basically, you should be able to go to settings and it should be the first one to be noted, but also it should wrap around. So if I scroll all the way down and I reach the bottom and then I keep going, I should go back to the top and the back button will be right there. So yes.

MarcaDian:
I have enabled Talkback, and I confirm that the problem is present, I will fix it later. Thanks for the feedback

MarcaDian:
With all due respect, I sympathize with you, I tried Talkback for the first time, it takes nerve, the navigation is so not obvious. The back button should be fixed in the next dev release soon.

LandonMoran:
I'm sorry for getting upset the first time. It works perfectly now. Thank you!

LandonMoran:
Good job on this! I'm glad it's here because when I went to use the manager there were some times that I got really upset because I have to use the settings sometimes which the entire settings menu in your credit is accessible but then getting out of it was impossible. So I would have to close the app and then reopen it and then do what I have to do and if I click settings on accident, oh no, I was stuck in there. Then I would have to close the app reopen it again but it works good now

LandonMoran:
One thing I found that's not a huge issue but is something is when you go to the sources menu. There's a close button at the bottom and a close button at the top cuz one of them says close and one of them says close sheet which is like which the close button already worked before this for that specific menu

MarcaDian:
I added closing the modal bottom sheet by clicking on the bottom sheet handler. I understand that your Talkback also works on an empty space, clicking on which also closes this bottom sheet. Because if your source list contains many sources, then there is no empty space, and only this handle remains. The documentation says that this is standard behavior, but if necessary, I can roll it back.

LandonMoran:
It's not an issue at all. I was just a little confused seeing two back buttons it works just fine

PR #586, opened 2026-05-25
fix: silence patcher foreground notification

### LINKED ISSUE #577: feat: Ability to disable patch complete sound or at least tie it to notification sound channel not ringtone.
### Feature description

I like having my phone only make noise for nessessary things like calls. But the recent addition of a notification noise when the patch is done even when the app is open is quite annoying, and it doesn't use the media or even notification audio channel so having the volume turned down doesn't help which I found out while patching an app before bed with my significant other almost asleep next to me.

Please make an option to disable the audio please

### Motivation

I have notification volume muted and system notifications muted, the only thing not muted is ringtone volume so important calls come through, and for some reason you chose the audio from this app to use ringtone so it's max volume every time unless I am willing to miss calls.

### Acknowledgements

ZaidNAlAsali:
## Summary

This PR marks the patcher foreground notification as silent.

The notification still uses the same channel, id, title/text, progress, pending intent, category, and ongoing behavior. The intended change is only that patcher notifications no longer alert with sound/vibration.

Closes #577.

## Testing

* Ran `git diff --check`
* Ran `.\gradlew.bat :app:compileDebugKotlin`
* Ran `.\gradlew.bat :app:assembleDebug`

### DIFF
--- a/app/src/main/java/app/morphe/manager/patcher/worker/PatcherWorker.kt
+++ b/app/src/main/java/app/morphe/manager/patcher/worker/PatcherWorker.kt
@@ -5,11 +5,11 @@ import android.app.*
 import android.content.Context
 import android.content.Intent
 import android.content.pm.ServiceInfo
-import android.graphics.drawable.Icon
 import android.os.Build
 import android.os.PowerManager
 import android.os.StatFs
 import android.util.Log
+import androidx.core.app.NotificationCompat
 import androidx.work.ForegroundInfo
 import androidx.work.WorkerParameters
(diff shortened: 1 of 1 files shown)


LisoUseInAIKyrios:
The patched sound was already [removed](https://github.com/MorpheApp/morphe-manager/commit/b0b5e9478418ab0e75d518d63c7ac30b0e33d003).

This change isn't needed.

ZaidNAlAsali:
Understood, thanks for clarifying. I’ll close this PR since it’s no longer needed.

PR #589, opened 2026-05-25
fix: Improve source import and installer reliability

openclawed-007:
## Changes

This PR tightens up a few related edge cases around adding patch sources, patching in mount mode, and cleaning up installer state.

- Source URLs now keep custom ports and nested GitHub/GitLab paths instead of being rewritten incorrectly.
- Failed local `.mpp` imports clean up after themselves, and settings import skips duplicate custom sources.
- Mount patching now stops early if filtering removes every selected patch.
- Manager update progress handles servers that ignore resume requests, and multipart bundle downloads fail if a range ends early.
- Temporary APK files, shared installer files, and PackageInstaller receivers are cleaned up in failure paths.
- Root mount metadata is written safely, signing files are excluded from backup, and settings exports no longer include the keystore password.

## Testing

- `git diff --check`
- `xmllint --noout app/src/main/res/xml/backup_rules.xml app/src/main/res/xml/data_extraction_rules.xml app/src/main/AndroidManifest.xml`
- `bash -n app/src/main/assets/root/service.sh`

I also set up local JDK 17 and tried `:app:compileDebugKotlin`, but this environment cannot access the Morphe GitHub Packages dependencies with the available token.

### DIFF
--- a/app/src/main/assets/root/module.prop
+++ b/app/src/main/assets/root/module.prop
@@ -1,6 +1,6 @@
-id=__PKG_NAME__-Morphe
-name=__LABEL__ Morphe
-version=__VERSION__
+id=__PKG_NAME_PROP__-Morphe
+name=__LABEL_PROP__ Morphe
+version=__VERSION_PROP__
 versionCode=0
 author=Morphe
 description=Mounts the patched APK on top of the original one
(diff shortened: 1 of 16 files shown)


LisoUseInAIKyrios:
Most of these changes are irrelevant/unnecessary.

Please avoid using OpenClaw without a clear goal of what to fix/improve.

openclawed-007:
Was not using openclaw for this, its just my account name

PR #590, opened 2026-05-25
fix: Escape root module metadata

OpenNerdz:
## Summary

Mount install writes APK metadata into the Magisk module files. The shell script template was using raw string replacement, so values like the APK version name could break out of the assignment in `service.sh`.

This escapes values based on where they are written: shell literals for `service.sh`, and newline/NUL-stripped values for `module.prop`.

## Changes

- Use shell-safe placeholders for `service.sh`.
- Sanitize `module.prop` values separately.
- Keep the change scoped to root module file generation.

## Testing

- `git diff --check`
- `bash -n app/src/main/assets/root/service.sh`
- Checked that a quoted version string with shell metacharacters stays inert after escaping.

### DIFF
--- a/app/src/main/assets/root/module.prop
+++ b/app/src/main/assets/root/module.prop
@@ -1,6 +1,6 @@
-id=__PKG_NAME__-Morphe
-name=__LABEL__ Morphe
-version=__VERSION__
+id=__PKG_NAME_PROP__-Morphe
+name=__LABEL_PROP__ Morphe
+version=__VERSION_PROP__
 versionCode=0
 author=Morphe
 description=Mounts the patched APK on top of the original one
--- a/app/src/main/assets/root/service.sh
+++ b/app/src/main/assets/root/service.sh
@@ -1,6 +1,6 @@
 #!/system/bin/sh
-package_name="__PKG_NAME__"
-version="__VERSION__"
+package_name=__PKG_NAME_SH__
+version=__VERSION_SH__
 
 # Resolve the module directory from the script's own path.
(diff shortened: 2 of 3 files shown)


LisoUseInAIKyrios:
This looks oddly identical to the completely different AI PR made just minutes before.

https://github.com/MorpheApp/morphe-manager/pull/589/changes

We should not be speculatively fixing the root script when we have not tested this and have not confirmed the speculative issue exists.

OpenNerdz:
I added a focused regression test in `RootInstallerTest` to make this concrete.

What it proves:
- The old generated shape, `version="<metadata>"`, allows shell syntax embedded in metadata to terminate the assignment and run a command.
- The new `shellLiteral` output keeps the same metadata inert.

Local repro output from the same assignment shape:

```
origin/main vulnerable template line: version="__VERSION__"
current fixed template line: version=__VERSION_SH__
unsafe generated assignment: version=""; printf vulnerable > "$PROOF"; #"
safe generated assignment: version='"; printf vulnerable > "$PROOF"; #'
unsafe marker exists: yes (vulnerable)
safe marker exists: no
```

Checks run:
- `git diff --check`
- `bash -n app/src/main/assets/root/service.sh`

I attempted the focused Gradle test with `./gradlew :app:testDebugUnitTest --tests app.morphe.manager.domain.installer.RootInstallerTest`, but this Termux environment cannot provide the project-required Java 17 toolchain. The test is committed on the branch so it can run in a normal Java 17/CI environment.

PR #598, opened 2026-05-28
fix: Cancelled install callbacks on some OEM systems

zjns:
On some OEM Android builds, such as ColorOS, PackageInstaller may report a cancelled result even after the installation has completed successfully.

This PR handles that compatibility case by checking the SHA-256 hash of the installed APK source after a cancelled callback. If the installed package body matches the APK that was just submitted for installation, the flow treats the install as successful; otherwise it preserves the existing cancel behavior.

Verification:
- ./gradlew :app:assembleDebug

### DIFF
--- a/app/src/main/java/app/morphe/manager/ui/viewmodel/InstallViewModel.kt
+++ b/app/src/main/java/app/morphe/manager/ui/viewmodel/InstallViewModel.kt
@@ -17,6 +17,7 @@ import app.morphe.manager.domain.installer.*
 import app.morphe.manager.domain.manager.PreferencesManager
 import app.morphe.manager.util.AppDataResolver
 import app.morphe.manager.util.PM
+import app.morphe.manager.util.sha256OrNull
 import app.morphe.manager.util.simpleMessage
 import app.morphe.manager.util.toast
 import kotlinx.coroutines.*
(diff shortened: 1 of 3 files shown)


LisoUseInAIKyrios:
We need to verify and thoroughly test because this may break other devices.

Does ColorOS have an emulator rom?

DonDae001:
@LisoUseInAIKyrios You can try using OxygenOS

OnePlus' OxygenOS is pretty similar to OPPO's ColorOS, as they're both under BBK Electronics and OPPO is the parent company of OnePlus

https://roms.danielspringer.at/index.php

zjns:
@LisoUseInAIKyrios I don't think large-scale testing is necessary. The current change does not introduce any scenarios where the installation status could be incorrectly detected, and this modification has been successfully tested on my ColorOS device.

MarcaDian:
@zjns Do my latest changes also work and didn't break anything? if yes then I'll merge it.

zjns:
> @zjns Do my latest changes also work and didn't break anything? if yes then I'll merge it.

It's working well and merge ok.

MarcaDian:
Ok. Thanks for the contribution 👍

PR #611, opened 2026-05-31
feat: Add onboarding tour triggered after first patch

### LINKED ISSUE #314: feat:  welcome tô morphe.
### Feature description

This feature request is about improving the first-time user experience with a simple onboarding guide inside the Manager.

When a user installs the Manager for the first time, a short “Welcome to Morphe” introduction could appear with a few screens explaining the basics of how the app works. This would help new users understand the core features without needing to immediately search for documentation or Reddit posts.

For example, something like this:

Screen 1 – Welcome
Welcome to Morphe.
You will see a few quick screens explaining how the Manager works and its main features. For more detailed help, you can always check the documentation and FAQ.

Screen 2 – How patching works
This application uses the Morphe Patcher and patch sources to modify applications.
You can customize supported apps as long as the required patches and sources are available.

Screen 3 – Advanced users
Advanced users who want more control and transparency can enable expert or advanced modes in the settings. These options allow deeper customization and more detailed information about what happens in the background.

Screen 4 – Faster updates
Users who want to receive updates earlier can enable pre-release or development updates for both patches and the Manager itself.

Screen 5 – Interface tips
You can long-press applications on the home screen to hide them if you want a cleaner interface.

The goal of this idea is simply to provide a small in-app guide that helps new users understand the main features quickly, instead of having to search through Reddit or external documentation every time they want to learn something.

### Motivation

The main motivation behind this idea is simple: everyone appreciates being welcomed and properly introduced to a project. A small onboarding flow can reduce user friction and make the first experience much more pleasant.

Another important reason is that this project relies heavily on user interaction. Making the experience clearer and more comfortable helps users understand the system and ultimately contributes to the overall health of the project.

Automation alone is not always the best approach. Sometimes improving the user experience and helping users understand the system is just as important.

The discussion is open, and if the team agrees with this direction, the idea is here to be considered.

### Acknowledgements

MarcaDian:
- Tour prompt dialog shown after the first successful patch install; appears after the notification permission dialog closes if one was shown
 - Multiphase coach-mark overlay across 13 steps in 3 phases: Home → Sources sheet → Settings
 - Spotlight cutout with pulsing animation highlights each UI element
-  Smooth fade-in entrance on overlay appearance
 - Animated card position - slides between top/bottom based on highlighted element location
 - Back button in the card and system Back key step backward through the current phase; reaching step 0 of a phase restores the previous screen or sheet before resuming the prior phase
 - Overlay is hidden during phase transitions to prevent ghost elements while navigating between screens
 - Each step automatically scrolls and navigates to its target — Settings tabs switch and scroll positions animate so the highlighted element is always in view
 - Sources sheet phase renders the overlay inside a separate Dialog window so it appears above the ModalBottomSheet's own popup layer
 - Coach-mark card is width-capped and position-clamped to remain fully on screen at any screen size and orientation
 - Swipe gesture hint stops immediately when advancing to the next step
 - Feature tour entry in Settings → System → About lets the user replay the tour at any time after the initial run
___


app/src/main/java/app/morphe/manager/ui/screen/home/OnboardingShowcase.kt
```diff
@@ -0,0 +1,246 @@
...
+import androidx.compose.ui.layout.boundsInWindow
+import androidx.compose.ui.layout.onGloballyPositioned
+import androidx.compose.ui.res.stringResource
+import androidx.compose.ui.text.font.FontWeight
+import androidx.compose.ui.text.style.TextAlign
+import androidx.compose.ui.unit.dp
+import app.morphe.manager.R
+
+/** Set to true to show onboarding every launch regardless of firstLaunch preference. */
+// TODO: Set to false before release
```
MarcaDian: TODO: Set to false before release

MarcaDian:
This may be reworked into a guide after the first patching.

PR #614, opened 2026-05-31
fix: Use existing credential helpers for gpr in settings.gradle.kts

### LINKED ISSUE #612: bug: GitHub Package Registry credentials in settings.gradle.kts ignore local.properties
### Bug description

Maven credentials for gpr in settings.gradle.kts only read from providers.gradleProperty() and System.getenv() ignoring local.properties

Helper functions githubUser() and githubToken() already exists that correctly resolve credentials from all three sources (local.properties -> gradle property -> environment variable) but the credentials block doesnt use them

This makes contributors with gpr.user / gpr.key set in local.properties get authentication failures when resolving dependencies.

1.20.0

### Morphe logs

```shell
n/a
```

### Acknowledgements

neon16bit:
Fixes #612

Maven credentials block for maven.pkg.github.com was bypassing local.properties, only checking gradle properties and environment variables. Helper fucntions githubUser() and githubToken() are defined at the top of the file already handle all three credential sources in the correct priority order: local.properties -> gradle property -> environment variable

Replaced the inline credential resolution with calls to the helpers functions, and removed the unused hardcodedUser/hardcodedToken variables.

### DIFF
--- a/settings.gradle.kts
+++ b/settings.gradle.kts
@@ -41,13 +41,8 @@ dependencyResolutionManagement {
             // A repository must be specified for some reason. "registry" is a dummy.
             url = uri("https://maven.pkg.github.com/MorpheApp/registry")
             credentials {
-                val hardcodedUser = ""
-                val hardcodedToken = ""
-                val gprUser: String? = providers.gradleProperty("gpr.user").orNull
-                val gprKey: String? = providers.gradleProperty("gpr.key").orNull
-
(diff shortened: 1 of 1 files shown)


LisoUseInAIKyrios:
Thanks for noticing, but the fix is already present in URV 1.6.0 which Morphe Manager was initially based off. We must have overlooked those few lines when importing URV.

https://github.com/Jman-Github/Universal-ReVanced-Manager/blob/597b3173a004f5a9aae54326046dd7fd4c5b7777/settings.gradle.kts#L51-L52

PR #629, opened 2026-06-11
Improve flavor text in strings.xml

Wackanim:
I was getting pretty sick of the boring, ChatGPT-ian loading messages, so I rewrote most of them to make them at least a bit more human sounding and ☝️maybe even a little funny.

### DIFF
--- a/app/src/main/res/values/strings.xml
+++ b/app/src/main/res/values/strings.xml
@@ -37,13 +37,11 @@
     <string name="home_all_apps_hidden_subtitle">You\'ve hidden all apps from the home screen</string>
 
     <!-- Home Screen - Home Greetings -->
-    <string name="home_greeting_1">What app do you want to patch?</string>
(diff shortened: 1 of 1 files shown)


MarcaDian:
These are not AI-generated sentences, even if there are emojis. Each sentence was chosen by us in a discussion.

Yes, now you have added a lot of human errors 😅, such as various quotes or sentences that start with a lowercase letter))

Keep in mind that sentences in toasts cannot be too long, otherwise they will be cut off by the system.

LisoUseInAIKyrios:
Many of these are slightly cringe, or may be confusing to user:
`Paying Hush Money to the Google lawyers...`
It sounds like we're involved in legal conspiracies.

I made an issue where anyone can post ideas. If any suggestions are amusing/useful we will add the suggestions to the app.

https://github.com/MorpheApp/morphe-manager/issues/630

PR #651, opened 2026-06-18
fix: Prevent text compression in headers on tablet portrait

kondratjev:
## Summary
Fixed layout issues where app name and step name text were being 
squeezed/compressed on tablet portrait (medium-width) screens.

The fix makes both layouts responsive: on wide screens (≥840dp) 
the original inline layout is preserved; on medium/narrow screens 
the secondary elements move below the primary text.

### `app/src/main/java/app/morphe/manager/ui/screen/home/InstalledAppInfoDialog.kt`
- Replaced hard-coded `compact = true` with `compact = windowSize.widthSizeClass == WindowWidthSizeClass.Expanded` when calling `AppHeroHeader` in the tablet (two-column) layout.
- **Effect:** the install-type chip and patch-date chip now move from the right side of the app name to a row below it on tablet portrait (600–840dp). The app name gets the full available width and is no longer compressed into a vertical letter-by-letter layout.

### `app/src/main/java/app/morphe/manager/ui/screen/patcher/ExpertPatchingProgress.kt`
- Added `maxLines = 1`, `overflow = TextOverflow.Ellipsis`, and `Modifier.weight(1f, fill = false)` to the "Patching app" title in `ExpertProgressHeader`.
- **Effect:** the percentage badge keeps its intrinsic width instead of being squeezed by the title text. Long localized titles are truncated with ellipsis rather than wrapping or breaking the layout.

<details>
  <summary>Screenshots before</summary>
  photo_2026-06-18 22 02 14
photo_2026-06-18 22 02 16  
</details>

<details>
  <summary>Screenshots after</summary>
photo_2026-06-18 22 02 12
photo_2026-06-18 22 02 18
</details>

### DIFF
--- a/app/src/main/java/app/morphe/manager/ui/screen/home/InstalledAppInfoDialog.kt
+++ b/app/src/main/java/app/morphe/manager/ui/screen/home/InstalledAppInfoDialog.kt
@@ -326,7 +326,7 @@ fun InstalledAppInfoDialog(
                                 packageName = packageName,
                                 installedApp = installedApp,
                                 accentColor = appAccentColor,
-                                compact = true,
+                                compact = windowSize.widthSizeClass == WindowWidthSizeClass.Expanded,
                                 modifier = Modifier.clip(RoundedCornerShape(bottomStart = 16.dp, bottomEnd = 16.dp))
                             )
                         }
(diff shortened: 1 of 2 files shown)


MarcaDian:
Oh, we didn't even notice this problem 😯

Thanks for the solution, if everything is fine we will merge it as soon as possible.

PR #666, opened 2026-06-27
fix: Don't trigger update badge for experimental-version-only changes

### LINKED ISSUE #622: bug: experimental version support triggers update badge in app
### Bug description

If patches update to support a new experimental version, and that's all that is down for that app, it will still trigger an update badge on the app list. This shouldn't happen unless there are updates to the actual patches instead, or a new recommended app version.

1.21-dev3

### Morphe logs

```shell
Attached

[morphe_logcat_2026-06-07-09_23.log.txt](https://github.com/user-attachments/files/28678044/morphe_logcat_2026-06-07-09_23.log.txt)
```

### Acknowledgements

Nickhasntlost:
Patch bundle changelogs add a scoped bullet like `Add experimental support for `X.Y.Z`` every time a new experimental app version target is added. hasChangesFor() previously counted any scoped bullet as a real change, so this alone would trigger the home screen update badge for users who aren't affected at all.

Track each scope's bullet text (not just the scope name) and ignore bullets that are purely an experimental version addition. A scope with at least one other bullet is still treated as changed.

Fixes #622

## Validation

Ran this logic against the real morphe-patches CHANGELOG.md history (318 releases): correctly suppresses the false-positive badge in 71 past cases that were experimental-only, while still flagging all 610 cases with a real change — 0 regressions introduced.

### DIFF
--- a/app/src/main/java/app/morphe/manager/util/ChangelogParser.kt
+++ b/app/src/main/java/app/morphe/manager/util/ChangelogParser.kt
@@ -1,13 +1,22 @@
+/*
+ * Copyright 2026 Morphe.
+ * https://github.com/MorpheApp/morphe-manager
+ */
+
 package app.morphe.manager.util
 
 /**
  * Represents a single version entry parsed from a CHANGELOG.md file.
+ *
+ * [scopedBullets] preserves the individual bullet bodies per scope so that
+ * [ChangelogParser.hasChangesFor] can distinguish substantive changes from
+ * bookkeeping bullets (e.g. "Add experimental support for X.Y.Z").
  */
 data class ChangelogEntry(
     val version: String,
     val date: String?,
     val content: String,
-    val affectedScopes: Set<String> = emptySet(),
+    val scopedBullets: Map<String, List<String>> = emptyMap(),
 )
 
 /**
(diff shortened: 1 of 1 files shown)


Nickhasntlost:
Hi, just wanted to check if this PR is on your radar. Happy to make any changes if needed!

LisoUseInAIKyrios:
But new expeirmental app versions are new if someone wants to use expeirmental.

The changes here add hard coded logic that rely on a very specific commit message, which patch bundles may not always follow. So then we will have new bug reports of "Ignore pre-release update setting doesn't work" when a commit message isn't in this exact format.

Nickhasntlost:
Thanks for the feedback, that's a valid concern. Open to suggestions on a better approach!

MarcaDian:
I've given this decision some further thought. It's worth trying this PR, and we need to respect the user experience even when it conflicts with the technical perspective.

A rebase onto current dev was necessary since 086f15e5 changed `hasChangesFor` to `take appNames: Collection<String>`. Otherwise this PR would either fail to compile or silently regress that feature.

Please review my changes to see if everything is okay.

Nickhasntlost:
Thanks for the rebase and the kind words! The changes look good to me.

PR #676, opened 2026-06-30
fix: Resolve TalkBack accessibility issues

### LINKED ISSUE #674: bug: neu accessibility bugs
### Bug description

# Accessibility Issues Report (Summary Included)

## Summary

This report covers three accessibility issues I found, along with one small accessibility improvement suggestion.

* The "Install automatically after patch completion" option does not expose its on/off state to screen readers.
* Navigation inside expanded patch sources causes the page to scroll instead of moving accessibility focus.
* During patching in Simple Mode, the screen reader becomes unresponsive on low-end devices (Moto G24 and Moto G35), while it works correctly on the Moto G75.
* Suggestion: add screen reader actions to reorder apps on the home screen.

## Issue 1 – "Install automatically after patch completion" state is not exposed

In **Advanced Settings**, the **"Install automatically after patch completion"** option does not announce its current state to screen readers.

When navigating to this control, there is no indication whether it is **enabled** or **disabled**, making it difficult to use the feature accessibly.

## Issue 2 – Accessibility navigation inside patch sources

After expanding a patch source (for example, **Official Package**), navigating forward with a screen reader should move the accessibility focus to the next elements.

Instead, the page scrolls, making it difficult to reach the following controls using accessibility navigation.

## Issue 3 – Screen reader becomes unresponsive during patching in Simple Mode

When running a patch using **Simple Mode**, the interface is blocked until the operation finishes.

On lower-end devices such as the **Moto G24** and **Moto G35**, the screen reader becomes unresponsive during the patching process, making it impossible to follow what is happening.

On the **Moto G75**, however, this issue does not occur. The interface remains responsive and the screen reader continues working normally throughout the patching process.

Because of this difference, I believe the issue is worth investigating, as it may be related to device performance or to how Simple Mode handles the patching process.

## Suggestion – Accessibility actions for app reordering

Now that the home screen supports reordering applications, it would be helpful to expose accessibility actions such as **Move forward** and **Move backward** for screen reader users.

I understand there may be technical limitations, so this is only a suggestion.

Thank you again for all the work on this project. Accessibility has improved significantly over the last few releases, and I'm happy to continue contributing by reporting issues and sharing feedback.

All

### Morphe logs

```shell
N/a
```

### Acknowledgements

MarcaDian:
Fixes the four accessibility issues:
1. Auto-install toggle now exposes on/off state to TalkBack.
2. Expanded patch source cards keep inner controls focusable by moving the click target onto the header only.
3. Simple Mode - progress is reported through a single polite live region throttled to 10% steps, and the constantly-updating progress widgets are silenced.
4. Apps on the home screen can be reordered with a screen reader via Move up / Move down accessibility actions.

### DIFF
--- a/app/src/main/java/app/morphe/manager/ui/screen/PatcherScreen.kt
+++ b/app/src/main/java/app/morphe/manager/ui/screen/PatcherScreen.kt
@@ -13,9 +13,13 @@ import androidx.activity.compose.BackHandler
 import androidx.activity.compose.rememberLauncherForActivityResult
 import androidx.activity.result.contract.ActivityResultContracts
 import androidx.compose.animation.AnimatedContent
+import androidx.compose.animation.EnterTransition
+import androidx.compose.animation.ExitTransition
 import androidx.compose.animation.core.FastOutSlowInEasing
 import androidx.compose.animation.core.animateFloatAsState
+import androidx.compose.animation.core.snap
 import androidx.compose.animation.core.tween
+import androidx.compose.animation.togetherWith
 import androidx.compose.foundation.layout.Column
 import androidx.compose.foundation.layout.fillMaxSize
 import androidx.compose.foundation.layout.fillMaxWidth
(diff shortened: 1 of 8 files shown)


optimatrix0:
Can't wait to test the improvements. I'm really, really excited!

MarcaDian:
https://github.com/MorpheApp/morphe-manager/issues/674#issuecomment-4844002513

optimatrix0:
I apologize for not looking into the issue sooner. That was my mistake. I know how annoying it is to have to repeat things, so thank you for just pointing out the error and handling it for us.

PR #681, opened 2026-06-30
feat: add Play Store installer modes

rushiranpise:
Adds Play Store installer options for patched APKs across non-root, root, and Shizuku flows.

The non-root path launches Android's package installer with EXTRA_INSTALLER_PACKAGE_NAME set to com.android.vending. The root path reuses Morphe's RootInstaller/libsu flow to run pm install -t -i com.android.vending -r for silent root installs. The Shizuku/Sui path reuses Morphe's silent Shizuku session installer with a Play Store installer package override.

Each path persists a distinct install type, appears in installer selection and app details, and keeps the existing rooted mount installer and normal Shizuku installer separate. Manager self-updates keep unsupported privileged Play Store installer paths blocked.

Inspired by KingInstaller's Play Store installer package behavior.

It's AI-generated, but I tested both non-root methods and they work as intended.

Unfortunately, I no longer have a rooted device, so I couldn't test the root method myself.


LisoUseInAIKyrios:
Is the purpose to fix apps that check for the installation source?

Using Google Play Store as the installation source will re-enable app store updates for the patched app. The user must then ignore Play Store updates or the update will revert the app back to unpatched .

If a specific patched app must have a specific installer source, why not patch the app to spoof the installer such as here https://github.com/MorpheApp/morphe-patches-library/blob/main/patch-library/src/main/kotlin/app/morphe/patches/all/misc/fix/changepackageinstaller/ChangePackageInstallerPatch.kt

Spoofing the installer source thru patching makes it idiot proof because the user doesn't need pick anything and they cannot select the wrong installer.

rushiranpise:
> Is the purpose to fix apps that check for the installation source?
> 
> Using Google Play Store as the installation source will re-enable app store updates for the patched app. The user must then ignore Play Store updates or the update will revert the app back to unpatched .
> 
> If a specific patched app must have a specific installer source, why not patch the app to spoof the installer such as here [MorpheApp/morphe-patches-library@`main`/patch-library/src/main/kotlin/app/morphe/patches/all/misc/fix/changepackageinstaller/ChangePackageInstallerPatch.kt](https://github.com/MorpheApp/morphe-patches-library/blob/main/patch-library/src/main/kotlin/app/morphe/patches/all/misc/fix/changepackageinstaller/ChangePackageInstallerPatch.kt?rgh-link-date=2026-07-01T05%3A50%3A56Z)
> 
> Spoofing the installer source thru patching makes it idiot proof because the user doesn't need pick anything and they cannot select the wrong installer.

Thanks, that makes sense.

I agree the patch you mentioned is the better fix when the patched app itself checks the installer.

This PR has a different use case.

For example, if I install a patched Waze APK normally, Android Auto does not recognize it as installed from Play Store, so Waze may not show up in Android Auto.

But if the patched Waze APK is installed using a Play Store-source installer flow, Android records the installer as Play Store, and Android Auto can see Waze as Play Store-installed.

So the patch metioned helps when Waze/Spotify/etc. checks itself. This installer flow helps when another app or system service, like Android Auto, checks Android’s saved install source.

MarcaDian:
I tried this PR and it seems to work; the badges show the installation source, though my system dialog doesn't. But if you say it works on Android Auto, then it's fine.

LandonMoran:
What exactly does this do

Figim:
> What exactly does this do

For example, on my Samsung phone (App about section) , I see the patched APK as installed from the Play Store.  This solves certain problems

PR #686, opened 2026-07-02
fix: Shizuku auto-install hang

### LINKED ISSUE #680: bug: Auto-install with Shizuku can get stuck indefinitely after patching
## Bug description

When using **Shizuku** as the selected installer with **Auto-install after patching** enabled, the installation process sometimes gets stuck indefinitely. The installer continues spinning, but the app is never installed, have to close manager app to cancel,

## Scenario 1: Patching the original APK

1. Select **Shizuku** as the installer.
2. Enable **Auto-install after patching**.
3. Patch an original APK using the **Installed app** option.
4. After patching completes, Morphe prompts to uninstall the existing app because of a certificate mismatch.
5. Tap **Uninstall/OK** on the system dialog.

### Expected behavior

The installation should begin only after the uninstall has completed successfully.

### Actual behavior

The installation process starts immediately after confirming the uninstall dialog, maybe before the app has actually finished uninstalling. The installer remains stuck spinning indefinitely, and the patched app is never installed.

## Scenario 2: Updating an already patched app

This also occur when patching an already patched app, where no uninstall is required. After patching completes, the installation spinner sometimes remains indefinitely and the app is never installed.

## Expected behavior

The patched APK should be installed automatically after patching completes.

latest dev and stable

### Morphe logs

```shell
na
```

### Acknowledgements

rushiranpise:
Fixes issue where Auto-install after patching could get stuck indefinitely when Shizuku was selected as the installer.

## What changed

- Resume the pending patched APK install after the user confirms uninstall for a signature mismatch.
- Ensure both patcher conflict paths continue the install after uninstall completes.
- Add a timeout around Shizuku PackageInstaller result callbacks so a missing callback cannot leave the UI spinning forever.
- If Shizuku reports failure, verify whether the patched APK was actually installed by comparing APK hashes before showing an error.

Fixes #680

It's AI-generated, but I tested with shizuku and it works as intended.

### DIFF
--- a/app/src/main/java/app/morphe/manager/domain/installer/SessionInstaller.kt
+++ b/app/src/main/java/app/morphe/manager/domain/installer/SessionInstaller.kt
@@ -21,6 +21,7 @@ import androidx.core.net.toUri
 import app.morphe.manager.R
 import app.morphe.manager.util.APK_MIMETYPE
 import app.morphe.manager.util.PLAY_STORE_INSTALLER_PACKAGE
+import kotlinx.coroutines.TimeoutCancellationException
 import kotlinx.coroutines.suspendCancellableCoroutine
 import rikka.shizuku.Shizuku
 import rikka.shizuku.ShizukuProvider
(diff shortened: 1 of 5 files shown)


MarcaDian:
Which model?) Opus?

rushiranpise:
> Which model?) Opus?

Sonnet. I’m too broke for Opus.

PR #689, opened 2026-07-04
feat: Multi-select bulk actions for saved APKs and patch selections

rushiranpise:
Adds multi-select bulk actions to the saved-APK management dialog and the patch-selection dialog, with a shared selection UX derived from the home multi-select bar. Includes the original per-type "Delete all" action and the refactoring needed to reuse selection UI across dialogs.
___

### DIFF
--- a/app/src/main/java/app/morphe/manager/ui/screen/home/ExpertModeDialog.kt
+++ b/app/src/main/java/app/morphe/manager/ui/screen/home/ExpertModeDialog.kt
@@ -912,13 +912,11 @@ private fun PatchOptionsDialog(
         onDismissRequest = onDismiss,
         title = patch.name,
         titleTrailingContent = {
-            IconButton(onClick = onReset) {
-                Icon(
-                    imageVector = Icons.Outlined.Restore,
-                    contentDescription = stringResource(R.string.reset),
-                    tint = LocalDialogTextColor.current
-                )
-            }
+            DialogTitleAction(
+                icon = Icons.Outlined.Restore,
+                contentDescription = stringResource(R.string.reset),
+                onClick = onReset
+            )
         },
         footer = {
             MorpheDialogOutlinedButton(
--- a/app/src/main/java/app/morphe/manager/ui/screen/home/HomeDialogs.kt
+++ b/app/src/main/java/app/morphe/manager/ui/screen/home/HomeDialogs.kt
@@ -76,7 +76,7 @@ fun HomeDialogs(
 
     // APK selection processing overlay - blocks interaction while APK is loaded/validated in background
     MorpheOverlay(visible = homeViewModel.processingApkSelection) {
-        PulsingLogoIndicator()
+        PulsingLogoWithCaption(caption = stringResource(R.string.processing_apk))
     }
 
     // Dialog 1: APK availability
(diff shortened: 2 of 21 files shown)


rushiranpise:
i have 72 apks in both sections. 💀

MarcaDian:
Add screenshot please.

MarcaDian:
I think we could also add the deletion of patched APKs, but without duplicating code or strings, except for the dialog header?

rushiranpise:
> I think we could also add the deletion of patched APKs, but without duplicating code or strings, except for the dialog header?

Done.

I refactored the delete-all flow so patched APKs and original APKs now share the same bulk delete confirmation UI and generic APK delete-all strings. Only the dialog titles remain separate for patched vs original APKs.

I also added multi-select support to the same shared APK management dialog:
- long press an APK tile to enter selection mode
- share selected APKs
- export selected APKs as a ZIP
- delete selected APKs
- select all
- selected total size is shown in the title

Batch ZIP exports use type-specific timestamped names, for example:
- morphe-patched-apks-YYYY-MM-DD-HHMMSS.zip
- morphe-original-apks-YYYY-MM-DD-HHMMSS.zip

I kept install as a single-item action for patched APKs only.

rushiranpise:
> Add screenshot please.

Screenshot_20260704-100745
Screenshot_20260704-100748
Screenshot_20260704-100756
<imgwidth="300" height="auto" alt="Screenshot_20260704-100822" src="https://github.com/user-attachments/assets/5a77f88c-ad21-422b-939d-77a6bf716c19" />
Screenshot_20260704-100838
Screenshot_20260704-100856

rushiranpise:
I moved the selected total size out of the dialog title because it could overflow beside the action icons. The title now stays compact as “N selected”, and the summary box switches to the selected APK count and selected total size while multi-select mode is active.

Screenshot_20260704-102406

MarcaDian:
The UI still needs some work; I think we could use the same ActionPill button style found in the card, adding it as a separate ActionRow at the top, below "selected".

Instead of checkboxes, a long press can be used, just as is done elsewhere in the manager.

I'll try to do this when I have time.

rushiranpise:
> The UI still needs some work; I think we could use the same ActionPill button style found in the card, adding it as a separate ActionRow at the top, below "selected".
> 
> Instead of checkboxes, a long press can be used, just as is done elsewhere in the manager.
> 
> I'll try to do this when I have time.

Done 😁

I removed the checkboxes and switched the APK list to long-press multi-select, matching the existing manager behavior.

The selected actions are now shown as an ActionPill row near the top instead of being placed in the dialog title. I also aligned the selection feedback with the home screen: selected APKs show an animated check badge, unselected APKs dim in selection mode, and long-press uses haptic feedback.

Delete all remains available only in the normal non-selection state.

Screenshot_20260704-105837

MarcaDian:
That's cool; the main thing is that the code is shared and doesn't duplicate what already exists.

MarcaDian:
I've made some changes; please check if everything is okay.

MarcaDian:
Note to myself: Exporting multiple files is not working correctly and does not reset the multi-selection state after export

MarcaDian:
It looks ready for release; the PR title and description need to be updated to reflect the changes.

rushiranpise:
> I've made some changes; please check if everything is okay.

Working good!  Thanks!  This would really help users with managing bulk apps like me 😆

Screenshot_20260705-110057

PR #715, opened 2026-07-12
feat: Route add-source through the website and enforce a remote blocklist

MarcaDian:
## Summary

  - The `https://morphe.software/add-source` intent-filter no longer uses `autoVerify`, so on Android 12+ bare URL clicks default to the browser and the website acts as the gateway.
  - New `BlocklistRepository` fetches a remote list from `https://api.morphe.software/v2/blocked-sources` on startup, caches it in DataStore for offline enforcement, and refreshes on every launch.
  - `PatchBundleRepository.createRemote()` refuses any add attempt whose repo appears on the blocklist. This covers deep links, manual URL entry, and share sheet, every path funnels through this chokepoint.
  - Blocked sources that were added before they entered the blocklist are retroactively filtered out of `enabledBundlesInfoFlow` and update predicates, and the sources management sheet shows a red "Blocked" badge with the reason.
  - A new shared `AlertSnackbar` surfaces four state-driven alerts on Home:
    - Blocked source detected
    - Updates paused on mobile data
    - Some patch sources are unavailable
    - Morphe update available

### DIFF
--- a/app/src/main/AndroidManifest.xml
+++ b/app/src/main/AndroidManifest.xml
@@ -112,8 +112,8 @@
                 <data android:mimeType="application/apks" />
             </intent-filter>
 
-            <!-- Deep link: App Links (https://morphe.software/add-source?github=...) -->
-            <intent-filter android:autoVerify="true">
+            <!-- Add-source deep link. Website acts as the gateway for repo validation -->
+            <intent-filter android:autoVerify="false">
                 <action android:name="android.intent.action.VIEW" />
                 <category android:name="android.intent.category.DEFAULT" />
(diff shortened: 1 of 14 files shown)


LandonMoran:
Is there malicious sources out there that you would want to block? Also, does this block manual adding of the sources? Like if you go and you click add and then you do a manual paste of the link

MarcaDian:
There haven't been any such cases so far, but we need to be ready to react quickly, and that requires this kind of functionality. After all, you can see how fast patches appear. A manager will be able to disable such bundles while displaying an appropriate message to the user.

PR #718, opened 2026-07-13
feat: Add launch update controls

rushiranpise:
Add Advanced > Updates toggles for checking Morphe app updates and refreshing patch sources when the Morphe app opens.

Keep both toggles enabled by default so existing launch behavior is preserved until the user opts out.

Leave manual update paths unchanged, including Home pull-to-refresh and per-source patch updates.

### DIFF
--- a/app/src/main/java/app/morphe/manager/ManagerApplication.kt
+++ b/app/src/main/java/app/morphe/manager/ManagerApplication.kt
@@ -123,7 +123,9 @@ class ManagerApplication : Application() {
         scope.launch(Dispatchers.Default) {
             with(patchBundleRepository) {
                 reload()
-                updateCheck()
+                if (prefs.updatePatchSourcesOnLaunch.get()) {
+                    updateCheck()
+                }
             }
         }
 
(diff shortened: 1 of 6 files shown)


LisoUseInAIKyrios:
There already is a toggle to disable updates for cellular, which is the only use case to not update.

I think adding niche settings, even in expert mode, isn't the solution to all problems .

rushiranpise:
> There already is a toggle to disable updates for cellular, which is the only use case to not update.
> 
> I think adding niche settings, even in expert mode, isn't the solution to all problems .

That's fair. My thinking wasn't specifically about reducing mobile data usage, but about giving users control over the automatic network requests on app launch. Some users may prefer faster startup, limited background activity, or working with a fixed set of patches until they choose to refresh.

The default behavior wouldn't change both toggles would stay enabled by default so this only affects users who explicitly opt out. Manual update paths (pull-to-refresh and per-source updates) would continue to work exactly as they do now.

LisoUseInAIKyrios:
Refreshing patch sources doesn't change app startup speed.

The refresh runs in the background, well after the app is already launched and available to use.

rushiranpise:
> Refreshing patch sources doesn't change app startup speed.
> 
> The refresh runs in the background, well after the app is already launched and available to use.

Got it, thanks for the clarification. That makes sense. I'll close the PR since this doesn't seem like the right fit.

PR #719, opened 2026-07-13
feat: Add universal patch default setting

rushiranpise:
Add an Advanced setting for controlling whether universal patches are automatically included in recommended/default patch selections.

When disabled, universal patches are skipped for simple-mode defaults, new default patch auto-selection, and Expert mode reset-to-default. Explicit saved selections and patcher-required dependencies are still respected.

https://github.com/MorpheApp/morphe-manager/issues/716

### DIFF
--- a/app/src/main/java/app/morphe/manager/domain/manager/PreferencesManager.kt
+++ b/app/src/main/java/app/morphe/manager/domain/manager/PreferencesManager.kt
@@ -65,6 +65,8 @@ class PreferencesManager(
 
     val useExpertMode = booleanPreference("use_expert_mode", false)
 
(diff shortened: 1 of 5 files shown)


LisoUseInAIKyrios:
I don't think this necessary.

Universal patches must be default off by default. It hasn't been enforced yet, but we can fix this in patcher and ignore declarations by third party devs that incorrectly set default on.

LisoUseInAIKyrios:
We will fix this in patcher, and ignore `default=true` for universal patches.

Universal patches must always per patch be opt in.

https://www.github.com/MorpheApp/morphe-patcher/issues/153

rushiranpise:
> We will fix this in patcher, and ignore `default=true` for universal patches.
> 
> Universal patches must always per patch be opt in.

Got it. Thanks 👍

PR #724, opened 2026-07-14
feat: optionally delete patched APK after install

### LINKED ISSUE #722: feat: do not store patched app after install
### Feature description

On devices with low memory, or in general, keeping an installed patched app separate in the app isn't entirely needed if they have the original apk and patch selection saved alongside it.

### Motivation

Saves space on devices with many patched apps.

### Acknowledgements

harrydbarnes:
## Summary

- add an opt-in **Delete patched APK after install** setting under Files & storage
- keep the original APK, patch selection, patch options, and installed-app metadata so the app can be repatched later
- keep installed apps visible in Morphe when their saved patched APK is deleted manually
- retain patched APKs for explicit saves and root-mounted installs, where the file is still required
- include the preference in Morphe settings import/export

## Why

Installed patched APKs do not need a second copy in Morphe when the original APK and patch configuration are already available. Skipping that copy can substantially reduce storage use for users with several patched apps.

Previously, deleting a patched APK through storage management also deleted its installed-app database record, making an app that was still installed disappear from Morphe's app list. Manual APK deletion now keeps that record whenever Android confirms the package is still installed.

Closes #722

## Validation

- `git diff --cached --check`
- parsed `app/src/main/res/values/strings.xml` as XML
- reviewed all successful install paths through the shared persistence callback
- verified home-list state is driven by the preserved `installed_app` record and live package-manager state
- APK installed and tested on a phone

### DIFF
--- a/app/src/main/java/app/morphe/manager/domain/manager/PreferencesManager.kt
+++ b/app/src/main/java/app/morphe/manager/domain/manager/PreferencesManager.kt
@@ -82,6 +82,7 @@ class PreferencesManager(
     val installerCustomComponents = stringSetPreference("installer_custom_components", emptySet())
     val installerHiddenComponents = stringSetPreference("installer_hidden_components", emptySet())
     val autoInstallWithShizuku = booleanPreference("auto_install_with_shizuku", false)
+    val deletePatchedApkAfterInstall = booleanPreference("delete_patched_apk_after_install", false)
 
     val useProcessRuntime = booleanPreference(
         "process_runtime", // Old key was 'use_process_runtime' and may have the wrong default for some devices.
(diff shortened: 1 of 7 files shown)


MarcaDian:
After reading your request, I initially planned to add two toggles to the Storage Management section, or a own dialogs, to ask whether to keep the saved/patched APK after patching (do not save). If this PR implements those changes, I think that would be great. I haven't looked at the code yet.

harrydbarnes:
> After reading your request, I initially planned to add two toggles to the Storage Management section, or a own dialogs, to ask whether to keep the saved/patched APK after patching (do not save). If this PR implements those changes, I think that would be great. I haven't looked at the code yet.

It adds a setting to the app that deletes the patched apk after install. Is that along the lines of what you were thinking? Could add a toggle on install also?

MarcaDian:
Thanks for your try, but we implemented this a bit more cleanly, simply by not keeping the APKs after patching.

harrydbarnes:
> Thanks for your try, but we implemented this a bit more cleanly, simply by not keeping the APKs after patching.

Thanks, saw. I went in and deleted the patched apks but now the app doesn't think I have anything patched. Not sure it's working fully

MarcaDian:
Manually removing the APK from the patched APKs dialog also deletes the database entry; everything should be fine the next time you patch.

PR #734, opened 2026-07-17
feat: Improve Shizuku installer flow

rushiranpise:
Add Shizuku/Sui silent uninstall support and use it for signature-conflict recovery when Shizuku auto-install explicitly opts into auto-uninstall. The install flow now waits for package removal before retrying and falls back to the existing conflict UI if silent uninstall cannot complete.

Expose a Shizuku status dialog from installer selection with installed/running/permission/mode details, permission requests, refresh, and open-Shizuku actions. Shizuku install failures now surface clearer messages while still preserving raw PackageInstaller details.

Collapse existing Play Store source variants into an Install as Play Store toggle for system, root, and Shizuku installers. The toggle keeps the existing PlayStore, RootPlayStore, and ShizukuPlayStore tokens underneath while warning users before enabling source spoofing behavior.


rushiranpise:
Screenshot_20260717-035714
Screenshot_20260717-035718
Screenshot_20260717-035735
Screenshot_20260717-035739

MarcaDian:
It looks cool visually. I'm currently working on a small redesign of the installer selection dialog, but I don't think there should be any conflicts.

I might not do it for now.

rushiranpise:
> It looks cool visually. I'm currently working on a small redesign of the installer selection dialog, but I don't think there should be any conflicts.
> 
> I might not do it for now.

Sounds good. I kept the changes mostly in the installer flow and reused the existing dialog components where possible, so hopefully it should not conflict much with your redesign.

MarcaDian:
I’ve finished redesigning the Installer section; do you need help resolving the conflict? It involves only visual changes.

rushiranpise:
> I’ve finished redesigning the Installer section; do you need help resolving the conflict? It involves only visual changes.

Thanks, I merged the latest dev and resolved the conflict. And adjusted the added toggles to use the new SettingsItem / MorpheSwitch style so they match the redesign.

rushiranpise:
Working as intended. Just that grant shizuku box is not aligned center 
Screenshot_20260717-140650

MarcaDian:
I'm fixing it now.

rushiranpise:
Thanks, looks better now

> I'm fixing it now.

MarcaDian:
Ready to merge?

rushiranpise:
> Ready to merge?

Yes

LandonMoran:
Okay this seems good and I think it is good. But the one problem I think there might be is when your app gets uninstalled. You lose all your settings. So if people don't know it's going to uninstall it'll just like go just lose all their data

rushiranpise:
> Okay this seems good and I think it is good. But the one problem I think there might be is when your app gets uninstalled. You lose all your settings. So if people don't know it's going to uninstall it'll just like go just lose all their data

This feature is off by default and completely optional. I think the bold warning should be enough to make users aware of what will happen before enabling it.

Also, Shizuku is mainly targeted toward power users, so I don't expect new users to enable this without understanding the implications. The main motivation behind this feature was to improve and speed up the patching process, not to change the default user experience.

PR #739, opened 2026-07-18
feat: Add batch app uninstall and reinstall actions

rushiranpise:
Add contextual multi-select actions on the home app list so selected installed apps can be uninstalled and saved patched apps can be reinstalled without leaving the selection flow.

Extend the saved original and patched APK storage dialogs with matching per-item and multi-select install, reinstall, and uninstall actions. Batch installs are queued through the existing InstallViewModel so installer selection, conflict handling, and unavailable-installer prompts stay shared.

Route explicit uninstall requests through InstallerManager so Shizuku, root, and system installer choices are respected, while mount installs continue to clean up the root module and saved record consistently.


rushiranpise:
Added one follow-up fix after testing:

- The app info action screen now refreshes after Mount/Unmount/Remount finishes, so the buttons update immediately without closing and reopening the screen.
- Also routed the Uninstall button on that same screen through the shared installer manager, so it respects the selected uninstall path like Shizuku/root/system, matching the new batch uninstall behavior.

MarcaDian:
I installed it for testing and noticed a few things: the "Select All" button should probably change to "Deselect All" when everything is selected. Also, if at least one of the selected apps is already installed, the "Install" button disappears from the multiaction bar.

rushiranpise:
> I installed it for testing and noticed a few things: the "Select All" button should probably change to "Deselect All" when everything is selected. Also, if at least one of the selected apps is already installed, the "Install" button disappears from the multiaction bar.

Thanks! I'll fix it. I can see that behavior in the Storage Management. Are you referring to that same behavior, or is this happening somewhere else?

MarcaDian:
After seeing this in the code yesterday, I initially thought it should be everywhere. It’s a dilemma: either show two buttons and sacrifice space (there are already 6+ buttons), or try replacing them with a single button, which compromises UX slightly, but is still better than having no button at all.

rushiranpise:
> After seeing this in the code yesterday, I initially thought it should be everywhere. It’s a dilemma: either show two buttons and sacrifice space (there are already 6+ buttons), or try replacing them with a single button, which compromises UX slightly, but is still better than having no button at all.

I agree, it's a bit of a tradeoff. The reason I changed it this way is because mixed selections made the existing actions misleading for example, showing Install/Reinstall or Uninstall when only some selected items support that action could lead to unexpected results.

In the latest commit, I kept the current single-action behavior but made it stricter: actions are only shown when the entire selection supports the same operation. This avoids invalid buttons while keeping the UI clean and avoiding adding more buttons to an already crowded layout.

For cases where selections have different capabilities, hiding the action felt safer than showing a button that only partially applies.

MarcaDian:
If click the button to uninstall the app(s) and then cancel the operation, "Unable to uninstall" errors/toast appear, likely due to a timeout.

rushiranpise:
> If click the button to uninstall the app(s) and then cancel the operation, "Unable to uninstall" errors/toast appear, likely due to a timeout.

Thanks,i missed this as i was using shizuku auto install/uninstall. 
Updated. Cancelled uninstall prompts now return immediately instead of timing out, and batch uninstall/install/reinstall queues keep moving after skipped, cancelled, failed, or conflicting items. Per-item errors are still shown, and the final summary toast now only includes non-zero counts, e.g. Uninstalled ,Skipped and  Reinstalled.

PR #740, opened 2026-07-19
feat: Add monochrome theme

rushiranpise:
Add a Monochrome theme preset with neutral light and dark color schemes, wired through the existing theme selection flow.

Use shared Monochrome surface helpers so settings groups, glass controls, sheets, and selection bars can quiet their surfaces without changing the existing layout rounding. Home app cards use the same card structure with solid Monochrome surfaces instead of app-color gradients.


rushiranpise:
Screenshot_20260719-123933

MarcaDian:
What about the InfoApp dialog (when tapping the app card)? Usually, the app's accent colors are used there in the header, the update banner, and the buttons. I’m not sure if that needs changing.

rushiranpise:
> What about the InfoApp dialog (when tapping the app card)? Usually, the app's accent colors are used there in the header, the update banner, and the buttons. I’m not sure if that needs changing.

Thanks, I updated the app info dialog. Normal themes still use the app accent colors, but Monochrome now routes those accent areas through the shared theme defaults, so the header, update/warning banners, and action buttons use the neutral Monochrome palette. Custom accent colors still work if the user chooses one.

Chiku-09:
woah
this looks so nice.

MarcaDian:
Yeah, I tried it out and it looks cool; there's just one thing I didn't get, the code contains light theme colors, but there isn't actually a light monochrome theme. Is it even necessary? If so, how should it be implemented, as a separate card or as a toggle? "Monochrome dark / light" is too long a name for cards.

RaceConditionWinner:
> Yeah, I tried it out and it looks cool; there's just one thing I didn't get, the code contains light theme colors, but there isn't actually a light monochrome theme. Is it even necessary?

Personally I liked the app card dark background more. 

But the colour of button as shown in screenshot looks counter intuitive or not that good.

It's better to have the previous default colour for the buttons as is. 


But the dark app card background looks a cool theme option to have

rushiranpise:
> Yeah, I tried it out and it looks cool; there's just one thing I didn't get, the code contains light theme colors, but there isn't actually a light monochrome theme. Is it even necessary? If so, how should it be implemented, as a separate card or as a toggle? "Monochrome dark / light" is too long a name for cards.

I changed this so light/dark/system stays as the main theme mode, and Material You / Monochrome are now separate color styles. That avoids long names like “Monochrome dark” and makes the light monochrome palette actually useful through Light + Monochrome. I also kept compatibility for the old monochrome theme value.

> But the colour of button as shown in screenshot looks counter intuitive or not that good.
> 
> It's better to have the previous default colour for the buttons as is.

That was custom accent color. this is deafult

Screenshot_20260720-012504

Screenshot_20260720-012510

RaceConditionWinner:
> That was custom accent color. this is deafult
> 

Ahh that makes sense 


This looks very good now

PR #777, opened 2026-07-25
feat: Add home app card color settings

rushiranpise:
Add an Appearance > Home screen setting for app card colors with default, custom gradient, and custom solid modes.

Keep default home cards on the existing app/source gradient behavior, falling back to Morphe's original gradient when an app has no bundle color, while custom gradient and solid modes intentionally override cards through the existing home card rendering pipeline.

Reset the app card color dialog back to Morphe's default gradient so users can return custom values to the original card palette before saving.

Hide the setting while Monochrome style is active because monochrome cards intentionally ignore gradient accents and use neutral theme surfaces.

Mirror custom color previews and glass placeholder highlights in RTL layouts so logical start and end colors render on the correct side.

Persist the color mode and selected colors through settings import/export, and share the frosted chip renderer between home cards and the settings preview.


MarcaDian:
I am more than certain that we don't need to store the three colors separately; it could be a single string containing all three (or even the solid color as well).

rushiranpise:
> I am more than certain that we don't need to store the three colors separately; it could be a single string containing all three (or even the solid color as well).

Good point. I updated this so custom app card colors are stored as a single encoded preference containing the gradient start/middle/end colors and the solid color. The encode/decode logic is centralized in AppCardColorDefaults, and settings, export/import, and home card rendering all reuse that shared path now.

PR #779, opened 2026-07-25
fix: Preserve mount installs for saved APK flows

### LINKED ISSUE #775: bug: Mount system is broken in App 1.24.3
### Bug description

https://github.com/user-attachments/assets/7c2f43e1-4b39-41ea-8374-9b0a76124cda

https://github.com/user-attachments/assets/e683ab04-b462-46f1-9c8f-6d4ce942d16b

[Morphe Log1 - Unmount-Mount.txt](https://github.com/user-attachments/files/30375326/Morphe.Log1.-.Unmount-Mount.txt)

[Morphe Log2 - Direct Remount.txt](https://github.com/user-attachments/files/30375327/Morphe.Log2.-.Direct.Remount.txt)

1.24.3

### Morphe logs

```shell
Using YouTube Music as the example because it doesn't take 5 minutes to re-patch. Affects all mounting instances.

Logs attached above with videos.
```

### Acknowledgements

rushiranpise:
Set the active install type before root mount installs so shared install-state listeners do not rewrite mounted apps back to the system installer type.

Route saved patched APK reinstalls through the mount installer when Mount is the selected primary installer, including app info, home batch reinstall, and patched APK storage flows. Other installer modes and original APK installs continue through the existing install queue.

Fixes #775

### DIFF
--- a/app/src/main/java/app/morphe/manager/data/room/apps/installed/InstalledApp.kt
+++ b/app/src/main/java/app/morphe/manager/data/room/apps/installed/InstalledApp.kt
@@ -42,3 +42,10 @@ data class InstalledApp(
     @ColumnInfo(name = "selection_payload") val selectionPayload: SelectionPayload? = null,
     @ColumnInfo(name = "patched_at") val patchedAt: Long? = null
 )
+
+/**
+ * Root mount binds the patched APK over the stock one, so it only applies while patching kept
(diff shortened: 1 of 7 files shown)


MarcaDian:
Have you checked the fix? Does it work?

rushiranpise:
> Have you checked the fix? Does it work?

yes, it was working.

MarcaDian:
OK. Thanks 👍

PR #780, opened 2026-07-25
feat: Add a shared list scrollbar across lists and dialogs

rushiranpise:
Adds a reusable scrollbar overlay with a draggable thumb, shown on the home app list, the sources sheet and every dialog with scrollable content. When a list is sorted by name it doubles as an alphabet fast scroll, showing the leading letter
while dragging. `MorpheDialog` now draws the scrollbar and the scroll-to-top button itself for its scrollable variant, and exposes its horizontal inset so dialogs that own a `LazyColumn` can anchor theirs to the true dialog edge.


Chiku-09:
I don't think this feature adds much value in its current form.

Personally, I've never felt the need for a scrollbar on the home screen. The page already contains a lot of UI elements ,buttons, app grouping, and other controls so adding another  element makes it feel a bit more cluttered.

From a UX perspective, it also doesnt solves a significant problem or noticeably improves navigation.

That said, this is just my personal opinion, and others may find it useful.

rushiranpise:
> I don't think this feature adds much value in its current form.
> 
> Personally, I've never felt the need for a scrollbar on the home screen. The page already contains a lot of UI elements ,buttons, app grouping, and other controls so adding another element makes it feel a bit more cluttered.
> 
> From a UX perspective, it also doesnt solves a significant problem or noticeably improves navigation.
> 
> That said, this is just my personal opinion, and others may find it useful.

I get your point, but I see it a bit differently.

A scrollbar is a pretty basic UI component that's present in almost every app with a scrollable list. It isn't really against good UI/UX, especially since it only appears while scrolling and hides itself when it's not being used, so it doesn't permanently add visual clutter.

For users with long app lists, it also provides a quick indication of their position in the list and makes navigation easier. It's a small quality-of-life feature rather than something intended to solve a major UX problem.

Of course, that's just my perspective, and it's fine if you don't find it useful.

rushiranpise:
Screenshot_20260727-213515

MarcaDian:
I understand that this change is justified if there are many applications. This current version has a bug in the patch sources where the list indicator stops halfway, and all the indicators move jerkily, as if jumping to the next value.

rushiranpise:
> I understand that this change is justified if there are many applications. This current version has a bug in the patch sources where the list indicator stops halfway, and all the indicators move jerkily, as if jumping to the next value.

Thanks! I just noticed both of those bugs were introduced after my last commit.

I  tried making the scroll stop at the first occurrence of each alphabet letter instead of scrolling continuously. That made it feel a bit more predictable, but it still needs some refinement.

MarcaDian:
Hold off on making changes for now; I'm still working on it.

MarcaDian:
Please test this changes.

rushiranpise:
> Please test this changes.

Looks better now. It still seems a bit jerky in A–Z/Z–A mode, but I think that pause is required to position the list correctly. 

However, I can't see the scrollbar in Settings.

MarcaDian:
Is a scrollbar needed there?

Also, I’m not sure if the A–Z delay behavior is correct, since I don’t usually use the scrollbar.

rushiranpise:
> Is a scrollbar needed there?
> 
> Also, I’m not sure if the A–Z delay behavior is correct, since I don’t usually use the scrollbar.

Yes, I think so. Users shouldn’t have any issues with this. Most apps, for example the Contacts app, also stop scrolling at the first occurrence of each alphabet in the list.

The scrollbar in the changelog looks good, so I think it would also be a good addition to other Settings sections if you agree.

MarcaDian:
This PR is more complex than it seemed at first glance. There are a lot of nuances and changes.

rushiranpise:
> This PR is more complex than it seemed at first glance. There are a lot of nuances and changes.

I tested the latest Cl, and everything looks perfect now. Sorry for the trouble, and thanks for all your work on this. I think it's ready to merge.

MarcaDian:
If any fixes are needed, we can add them in separate PRs. Thanks for the idea and the implementation; I’ve wanted to add scrollbars for a long time.

PR #791, opened 2026-07-30
fix: Correct app status, signature and saved APK checks

### LINKED ISSUE #786: bug: Installed app status can become stale or incorrect
### Bug description

I have run into three cases where the status shown by Morphe does not match the actual state of the app or its saved APK.

The first case happens when "Keep patched APKs" is disabled:

1. Disable "Keep patched APKs" in Morphe.
2. Patch and install an app.
3. Uninstall that app from Android settings.
4. Reopen Morphe.

The app is no longer installed and there is no saved patched APK, but Morphe continues to treat its entry as an installed app. I would expect Morphe to clear the stale tracking data and return the entry to its normal uninstalled state.

The deletion check currently requires a saved copy to exist ([PM.kt](https://github.com/MorpheApp/morphe-manager/blob/80fff59400d740469c7fecc31fcc40308eef419e/app/src/main/java/app/morphe/manager/util/PM.kt#L211-L214)), and the home screen passes the result of the saved-file lookup to that check ([HomeViewModel.kt](https://github.com/MorpheApp/morphe-manager/blob/80fff59400d740469c7fecc31fcc40308eef419e/app/src/main/java/app/morphe/manager/ui/viewmodel/HomeViewModel.kt#L1251-L1258)).

The second case is an installed app being reported as patched when Morphe cannot read its signature. A failed signature check should not be treated as confirmation that the installed app is patched.

The signature helper returns an empty set when it cannot read the signatures ([PM.kt](https://github.com/MorpheApp/morphe-manager/blob/80fff59400d740469c7fecc31fcc40308eef419e/app/src/main/java/app/morphe/manager/util/PM.kt#L151-L163)). The metadata check then uses `none` on that result, which also returns true for an empty set ([HomeViewModel.kt](https://github.com/MorpheApp/morphe-manager/blob/80fff59400d740469c7fecc31fcc40308eef419e/app/src/main/java/app/morphe/manager/ui/viewmodel/HomeViewModel.kt#L1923-L1928)).

The third case concerns saved APKs. Morphe treats a file at the expected path as a saved copy even when that file is empty, corrupt, or belongs to a different version. This can leave the app shown as having a saved copy even though the APK cannot actually be used.

The saved-copy lookup only checks whether a file exists at either expected path ([HomeViewModel.kt](https://github.com/MorpheApp/morphe-manager/blob/80fff59400d740469c7fecc31fcc40308eef419e/app/src/main/java/app/morphe/manager/ui/viewmodel/HomeViewModel.kt#L1633-L1637)).

Morphe Manager v1.24.3

The issues were found on `dev` at [`80fff594`](https://github.com/MorpheApp/morphe-manager/commit/80fff59400d740469c7fecc31fcc40308eef419e).

This is not tied to a specific patched app or app version. The three cases concern how Morphe tracks installed apps and saved APKs.

### Morphe logs

I deliberately did not include logs because the issue concerns Morphe's state handling in the three cases described above, rather than a patching failure for a particular app.

### Acknowledgements

Nickhasntlost:
## Problem

Three separate cases where Morphe's displayed app status does not match the actual device state:

1. Uninstalled app stays shown as installed (Keep patched APKs disabled)
`isAppDeleted()` required `hasSavedCopy` to be true before returning true. Apps uninstalled while "Keep patched APKs" was disabled were never detected as deleted, leaving a stale installed status on the home screen.

2. Unreadable signatures treated as confirmed patched
`getInstalledSignatureHashes()` returns an empty set when signatures cannot be read. Calling `.none{}` on an empty set always returns `true`, which incorrectly marked the app as patched whenever signature reading failed. Now falls through to DB tracking instead.

3. Empty/corrupt saved APK treated as valid
`savedPatchedApkFile()` only checked whether a file existed at the expected path. A zero-byte or corrupt file was counted as a valid saved copy. Now also checks `file.length() > 0`.

## Fix

- `PM.kt` — removed `&& hasSavedCopy` from `isAppDeleted()` return statement
- `HomeViewModel.kt` — added empty-set guard before `.none{}` signature check
- `HomeViewModel.kt` — added `&& it.length() > 0` to `savedPatchedApkFile()`

## Validation

Built successfully with `assembleDebug` (50 tasks, 0 errors).

Fixes #786

### DIFF
--- a/app/src/main/java/app/morphe/manager/ui/viewmodel/HomeViewModel.kt
+++ b/app/src/main/java/app/morphe/manager/ui/viewmodel/HomeViewModel.kt
@@ -1634,7 +1634,7 @@ class HomeViewModel(
         listOf(
             filesystem.getPatchedAppFile(app.currentPackageName, app.version),
             filesystem.getPatchedAppFile(app.originalPackageName, app.version)
-        ).distinctBy { it.absolutePath }.firstOrNull { it.exists() }
+        ).distinctBy { it.absolutePath }.firstOrNull { it.exists() && it.length() > 0 }
 
     suspend fun persistReinstalledApp(
         app: InstalledApp,
@@ -1922,7 +1922,14 @@ class HomeViewModel(
                 }
                 val expectedSignatures = bundleAppMetadataFlow.value[packageName]?.signatures
                 if (!expectedSignatures.isNullOrEmpty()) {
-                    pm.getInstalledSignatureHashes(packageName).none { it in expectedSignatures }
+                    val installedHashes = pm.getInstalledSignatureHashes(packageName)
+                    if (installedHashes.isEmpty()) {
+                        // Cannot read installed signatures → skip verification, don't treat as patched
+                        val trackedPatch = installedAppRepository.get(packageName)
+                        trackedPatch != null && pkgInfo.versionName == trackedPatch.version
+                    } else {
+                        installedHashes.none { it in expectedSignatures }
+                    }
                 } else {
                     val trackedPatch = installedAppRepository.get(packageName)
                     trackedPatch != null && pkgInfo.versionName == trackedPatch.version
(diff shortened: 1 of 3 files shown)


MarcaDian:
Thanks 👍

Nickhasntlost:
You're welcome!

PR #795, opened 2026-07-31
feat: Patch several apps in one queue, and keep them patched automatically

MarcaDian:
## Description

Patching two or three apps used to mean going through the whole flow separately for each one, answering the same questions every time. Now user can hand Morphe a list and walk away.

### Patching several apps at once

Long-press an app on the home screen to start selecting, tap the others user want, then tap the wand button.
Morphe shows what it is about to do: which APK it found for each app, how many patches it will apply, and whether anything is missing.
Apps that need attention are marked, so user can attach a file or drop them from the list before anything starts.

Then it works through them one by one. The patching screen looks exactly like it always does, with a counter on top telling user how far along the queue is. User can leave the screen, it keeps going in the background.

When it finishes user get a summary and can install everything at once, or pick apps one at a time. If an install fails, the app says why right on its card, and you can try again without patching it a second time.

### Keeping apps patched by itself

There is a new option in `Settings` that lets Morphe re-patch apps in the background whenever their patches get updated, and quietly tell user when they are ready to install.
It is off by default, and it only touches apps it can rebuild on its own, without asking user anything.

### Launcher shortcuts

Long-press the Morphe icon for quick access to re-patching outdated apps, checking for updates, or jumping straight into an app user patched recently.

Full walkthrough: [Patching several apps at once](https://github.com/MorpheApp/morphe-manager/blob/feat/batch-patching/docs/batch-patching.md)
___


optimatrix0:
I have one question about the UI. How do you envision the app selection workflow working?

I'm wondering how apps will be selected and added to the queue. Maybe an "Add to queue" action in the actions/context menu would work well, but I'm curious to see what you have in mind. I'm also wondering where the queue button will be placed, since I'd like to make sure the whole workflow is easy to navigate.

If you'd like feedback, I'd be happy to test it and report any accessibility issues I find.

Congratulations on all the hard work and dedication you've put into this feature. I'm really excited to see it coming together and can't wait to try it! 🎉

optimatrix0:
I saw the discussion about this feature, and I'd like to offer my help with its accessibility from the very beginning.

If you'd like, I'd be happy to test it, provide feedback, and report any accessibility issues I come across. I'm looking forward to helping make this feature as polished and accessible as possible.

optimatrix0:
I just tested it, and it looks great overall. Everything seems to be working as expected.

The only suggestion I have is that it might be helpful to expose a few actions through the accessibility actions menu, such as selecting items. Other than that, the "Patch selected" button works well, and the overall workflow feels good. Nice work!

optimatrix0:
The only bug I found was with the "Install all" button.

The first app in the queue, YouTube, was installed successfully. However, the second app I added to the queue, YouTube Music, did not install. So this seems to be the only issue I found with the queue workflow.

I haven't tested the automatic repeat feature yet, but I configured it to run once a day and I'll keep monitoring it to provide more feedback and help improve the feature.

optimatrix0:
I noticed that the automatic installation button sometimes does not work, but I couldn't identify the exact reason.

For example, if an app like TikTok has patches from two different sources (such as the Morphe patch source with Disable Play Store Update and the TikTok Patches source), it seems to select only one of the sources, in this case the Morphe source.

It looks like the issue may be related to handling multiple patch sources during the automatic installation workflow.

Another suggestion for improving the workflow: adding a "Select patches" button next to each app, alongside the option to add it to the queue, could be very useful.

This way, users can decide exactly which patches they want to apply to each app before creating the queue. Since the user is the one choosing what modifications will be made, having this control would make the process more flexible and predictable.

These are just suggestions to help mature the feature and improve the overall user experience.

MarcaDian:
Please refrain from posting "walls of text." I appreciate your good intentions, but as a non-native speaker, I find them difficult to read. Your comments have been noted and will be implemented.

optimatrix0:
If the language barrier is making communication difficult, please let me know your native language. I'd be happy to translate my feedback so it's easier to discuss these ideas.

I also have one small suggestion: it would be nice if queue processing had completion notifications, just like a normal patch. For example, if the queue contains Instagram and X, I'd expect a notification when Instagram finishes, and another one when X finishes. That would make it much easier to keep track of the progress without having to constantly check the app.

The remaining bugs and refinements can always be addressed after the initial release. I'm just happy to help however I can, and I'm glad my feedback has been useful. Thanks again for all the work you've put into this feature!

LandonMoran:
I like all of the things about this feature, but one thing I wish would be added is when you get to the screen where you can pick to install them. It should show you which ones you have installed and which ones you haven't because in the current implementation you can install an app twice and you don't know which ones have installed or not unless you back out. So I would prefer that it disables the install button for ones that have already been installed when the auto install with shizuku is not set. And also it should just generally not let you install something again that you have already installed

ankitvrsharma:
It would be better if trigger to automatic re patching include "when patch source updates".

MarcaDian:
> I like all of the things about this feature, but one thing I wish would be added is when you get to the screen where you can pick to install them. It should show you which ones you have installed and which ones you haven't because in the current implementation you can install an app twice and you don't know which ones have installed or not unless you back out. So I would prefer that it disables the install button for ones that have already been installed when the auto install with shizuku is not set. And also it should just generally not let you install something again that you have already installed

It is difficult to understand what is written here, but if the app is already installed, it will simply update; if it isn't installed, it will be installed. Where is the problem? If you are modifying the package via a patch, then having two apps with different packages installed is also the expected behavior.

LandonMoran:
Okay let me explain it a little bit better. So when you batch select apps to patch they all go through then show you that you can install them if you install one it successfully installs then you're back on the screen. You can click the install button again and it will override the one you just installed with the same one. It'll just reinstall it over itself. You should not be able to do that. If you have installed the one of the listed apps of the list you just patched it should not let you install it again

MarcaDian:
> You should not be able to do that. If you have installed the one of the listed apps of the list you just patched it should not let you install it again

Why? What is the reason? I don't see a problem with that. This doesn't affect anything.

LandonMoran:
Because if you just patched like six apps at once, I have a little bit of trouble keeping track of what I have clicked installed installed on and what I haven't yet. It's not necessarily a bad thing it's just A user experience thing

MarcaDian:
> Because if you just patched like six apps at once, I have a little bit of trouble keeping track of what I have clicked installed installed on and what I haven't yet.

Now that makes sense. We'll try to fix it soon.

LandonMoran:
> It would be better if trigger to automatic re patching include "when patch source updates".

I don't think that's possible. I don't think we have the ability to in the backgroun detect when a source updates except for the official one

ankitvrsharma:
It could be possible as morphe manager keeps record of when the given patch source is updated.

LandonMoran:
> It could be possible as morphe manager keeps record of when the given patch source is updated. 

Yes, but that record only updates when the app is opened and forced to fetch the new patches which means that you wouldn't be able to get it from the background technically we could have morphe manager use a GitHub token but that would only allow it to get patch updates from sources that are on GitHub, which is a fair bit but not every single one

PrimePax:
Add Save/export Apk Option after batch patching Successful , Now there is only one option which is to install it.

MarcaDian:
> Add Save/export Apk Option after batch patching Successful , Now there is only one option which is to install it.

By default, the patched app is saved in the patched APKs directory; you can install it via the corresponding menu in the settings. This is mentioned at the top.

MarcaDian:
Please, if there are any bugs or (reasonable) suggestions, create separate issues, as it is difficult to track them across different places.

dr-strange15:
Thx big marca
You along with other developer are doing great work 😃


Also will this batch patching and automatic install feature be added in the morphe desktop too?

PR #797, opened 2026-07-31
feat: Add APK download helper integration

rushiranpise:
Expose an optional intent contract for third-party APK download helpers so Morphe can hand off original APK requests with package, version, build code, ABI, file type, and stock-install context.

Show the helper action only when Android can resolve a compatible helper, launch it from the existing download instructions dialog, and consume the returned APK-family URI through the existing APK selection flow.


LandonMoran:
So what does this do? Does this allow you to download apks as inside the app

rushiranpise:
> So what does this do? Does this allow you to download apks as inside the app

Not exactly. It doesn't download APKs itself. Instead, it provides an intent that allows third-party apps to download APKs for Morphe

LandonMoran:
Is there any apps that is going to support the set launch

dr-strange15:
> > So what does this do? Does this allow you to download apks as inside the app
> 
> Not exactly. It doesn't download APKs itself. Instead, it provides an intent that allows third-party apps to download APKs for Morphe

there must me some reasons for not doing this

but why not let morphe download the apks itself??
i mean this would remove so many manual steps

MarcaDian:
Users don't look at the bigger picture. Need I remind you what happened to Vanced? Adding things like this could be a mistake, so the pros and cons need to be carefully weighed.

dr-strange15:
> Users don't look at the bigger picture. Need I remind you what happened to Vanced? Adding things like this could be a mistake, so the pros and cons need to be carefully weighed.

yes yes big marca
thats why i said there must be some reasons

pardon me🙏

LisoUseInAIKyrios (commented):
Can the code here validate that `Intent.FLAG_GRANT_READ_URI_PERMISSION` is present and if not then show a simple error message?

This would be the fault of the external app. If some kind of error is shown then this is ok as-is.

MarcaDian:
https://github.com/user-attachments/assets/b9994576-b4a2-4f6f-817f-e4f46ebc68a1

LandonMoran:
What app supports this?

MarcaDian:
https://github.com/rushiranpise/apk-download-helper

PR #829, opened 2026-08-08
feat: Filter the home app list by patch and install state

rushiranpise:
Screenshot_20260808-163804
Screenshot_20260808-163755

- Add a Filter tab to the home app list sort dialog so users can narrow the home list by patched, not patched, installed, or uninstalled state without adding another home navigation control.

- Apply the selected filter before All apps, Sources, and Custom grouping so grouped views only show matching apps, and show a clear-filter empty state when nothing matches.

- Keep the existing sort entry point for the combined Sort and Filter dialog, highlight it when a filter is active, and update the Appearance setting label to match.

- Remove the old unused Sort apps dialog title and stale localized sort-button strings so changed labels fall back to the base wording until translations are updated.


DonDae001:
For UI consistency, the sort and filter buttons could be similar to the remote and local buttons in the "Add patch source" screen:

Screenshot_20260809_180922_Morphe

MarcaDian:
There is no need to touch the translation files, as they are always overwritten from Crowdin.

rushiranpise:
Looks better now 👍

Screenshot_20260810_134452_Morphe Debug

TripleEYE:
Just installed this dev version. Looking really good. Thought I'd give the positive feedback.
This filter idea is wonderful 👍

Also, I don't know if this is the intended result or not, but the installed filter also shows apps that are installed in their unpatched version but are not, and were never, patched by Morphe. For example, I have Gboard but I never used Morphe on it even once. Perhaps this is on purpose, I don't know; but it surprised me so I figured I'd point it out.

MarcaDian:
"Installed" refers to those installed on the device for which patches are available. If I’ve understood correctly how this is supposed to work

TripleEYE:
> for which patches are available.

If that's the case, I think this is working perfectly 👍

rushiranpise:
> Yes, "Installed" refers to apps that are installed on the device and have available patches (patchable + patched apps).

The idea behind this feature was to make managing a large number of apps easier. I have many apps installed, and manually finding them in a long list is time-consuming. Previously, using "Select All" would literally select everything. Now, you can use "Select All" and it will only select the installed/ patchable apps, making actions like uninstalling or managing them much faster.

TripleEYE:
I understand it now 😀 . I think this is a great idea! ❤️
The whole point of the Sort suggestion to begin with was easier organization and an easier time finding and managing our apps & patches. 
And this filter addition improves this very much 👏⭐⭐⭐⭐⭐👏

Edit: I'm just saying this from a user perspective, of course 😅

DonDae001:
Hi @rushiranpise while you're still active: Is it possible to make filtering persist even after Morphe fully closes? Sorting already does this.

The idea is to (finally) have a way to hide the new patchable apps that appear every time a patch source updates, and only show the apps I have patched.

rushiranpise:
> Hi @rushiranpise while you're still active: Is it possible to make filtering persist even after Morphe fully closes? Sorting already does this.
> 
> The idea is to (finally) have a way to hide the new patchable apps that appear every time a patch source updates, and only show the apps I have patched.

Initially, this was the default behavior. However, after looking at this [issue/comment](https://github.com/MorpheApp/morphe-manager/issues/820#issuecomment-5216476844), the same situation could happen again: users might forget that a filter is active and then panic, thinking, "Oh no, I can't find XYZ app."

To avoid this confusion, the current behavior keeps the filter only for the current session. Once the app is closed and launched again, the filter is cleared, so users always see the complete list by default.

TripleEYE:
Perhaps that can be resolved by making the sort/filter main menu button colorful when a filter is active. Example:
IMG_20260811_092057

And/or adding a long-press function to that same button to "clear all filters". Though I'm not sure if a long-press fits the manager ux aesthetics or not.

DonDae001:
Or, it could be a setting instead. There's already a toggle-able option in Settings > Appearance for the sort & filter button, perhaps it could be expanded to have an option to persist sort or filter settings after relaunch.

With this approach, the default values can stay the same (sort persists; filter doesn't) but it can be configurable.

PR #831, opened 2026-08-09
fix: Validate saved APKs and tracked installs

### LINKED ISSUE #830: bug: Stock reinstall remains marked as the tracked patched app
### Bug description

After updating to Morphe Manager v1.25.0, I can reproduce a remaining installed-state problem on a real device.

#### Steps to reproduce

1. Patch Ampere with Morphe and install the patched build.
2. Uninstall Ampere outside Morphe.
3. Install the stock Ampere build from Google Play.
4. Reopen or refresh Morphe.

#### Actual behavior

Ampere still appears in Morphe as an installed tracked app, even though the currently installed package is the fresh Google Play build. Because the package name is unchanged (`com.gombosdev.ampere`), package presence is being treated as if it proved that the tracked patched build is still installed.

This can also let Morphe reconcile its retained tracking row to the stock version and expose launch or uninstall actions against the stock app.

#### Expected behavior

Morphe should distinguish the tracked patched artifact from a stock package installed later under the same package name. A confirmed stock replacement should not be shown as the tracked patched install and should not receive destructive installed-app actions. If the available evidence cannot identify the build, Morphe should show an explicit unverified state instead of assuming it is patched or stock.

#### Why #791 does not cover this case

[#791](https://github.com/MorpheApp/morphe-manager/pull/791), which resolved [#786](https://github.com/MorpheApp/morphe-manager/issues/786), handles the missing-package case. Here the package is present again, but it is a different installation. The persisted Morphe record survives the uninstall, so checking only the package name cannot establish that the installed build is the one Morphe tracked.

#### Proposed patch 1: validate retained patched APKs

This is the narrow follow-up already proposed while reviewing #791: [the `isUsableApk()` patch in resolved issue #786](https://github.com/MorpheApp/morphe-manager/issues/786#issuecomment-5181652473).

Instead of accepting any non-empty file at the expected path, require a regular, parseable, signed APK whose package name and version match the tracked app. Apply the same validation in both the home-screen and installed-app-info saved-copy lookups. This prevents corrupt, wrong-package, or wrong-version files from enabling saved-APK actions.

#### Proposed patch 2: verify tracked-install identity

Resolve the installed state using artifact identity rather than the database row or package-name presence:

- compare the installed signing certificate with a valid retained patched APK when available;
- compare it with a retained original APK or bundle to identify a stock replacement;
- retain the mounted-source signal and installer identity as narrower fallback evidence;
- use an explicit unverified state when signatures or reference artifacts are unavailable;
- only expose launch/uninstall and version reconciliation for a verified tracked patched install.

The two patches address the same underlying assumption: a path or package being present does not prove that it is the artifact Morphe previously tracked.

- Morphe Manager v1.25.0 (`versionCode=39777118`)
- Ampere v4.36.1 (`com.gombosdev.ampere`, `versionCode=291`)
- Pixel 9

### Morphe logs

```shell
$ adb shell dumpsys package app.morphe.manager
Package [app.morphe.manager]
  versionCode=39777118
  versionName=1.25.0
  lastUpdateTime=2026-08-06 15:19:16

$ adb shell dumpsys package com.gombosdev.ampere
Package [com.gombosdev.ampere]
  versionCode=291
  versionName=v4.36.1
  installerPackageName=com.android.vending
  firstInstallTime=2026-08-09 01:18:50
  lastUpdateTime=2026-08-09 01:18:50
```

The package-manager state confirms that the currently installed Ampere is a fresh Google Play installation. Morphe nevertheless continues to display its retained patched-app record as installed. No patching failure is involved.

### Acknowledgements

- [x] I have checked all open and closed bug reports and this is not a duplicate. This is a device reproduction on v1.25.0 of the residual same-package replacement case left after #791.

AndreaCovelli:
## Summary

- validate retained patched APKs before offering install, export, or mount actions
- distinguish a tracked patched install from a stock app reinstalled under the same package name
- represent unverifiable installs explicitly and avoid destructive actions against them
- keep the added identity checks off the home screen's critical path

Fixes #830. Related to #786 and follows up on #791.

## Root cause

#791 handles the case where the tracked package is absent. The installed-app database row survives an uninstall, though, so a later stock reinstall under the same package name was still treated as the tracked patched build. That could make Morphe show the app as installed, reconcile the stale row to the stock version, and offer launch or uninstall actions for the stock app.

This was reproduced with Ampere: the patched build was uninstalled and the Play Store build was installed shortly afterward. ADB confirmed that the current package was installed by `com.android.vending`, while Morphe retained the old tracked record.

### Identifying the tracked install

- Resolve tracked-install state from evidence rather than from package-name presence or the persisted row, in this order: a mounted-source signature mismatch, the saved patched APK's certificate, the retained original or bundle certificates, a patch-manager installer, and finally an installer that Morphe would not have set for the record combined with an installation newer than the record itself.
- The installer signals are read against the record's `installType`: Play Store-attributed, custom and mounted installs carry an attribution Morphe cannot rule anything out from, while the system-installer and Shizuku modes have a predictable one. This is what identifies the reported Ampere case without any archive being retained.
- Require a saved APK to be a regular, parseable, signed APK whose embedded package and version match the record. Both patch flows persist the version the patcher produced and reuse it in the retained file name, so version-rewriting patches still match their own record while an archive that reports something else is refused. The expected file name is not evidence of what the file contains, and the accepted archive goes on to supply the certificate that proves an install is patched.
- For renamed apps, search both retained-file paths but accept only an APK whose embedded package matches the current tracked package.

### Acting on it

- Treat a confirmed stock replacement as deleted from Morphe's perspective while leaving the stock app untouched; the stale tracking record can still be removed.
- Show an `Unverified` state when the available evidence cannot identify the installed build, suppress update badges, and withhold launch/uninstall actions. The notice explains the state and points at reinstalling from the saved APK.
- Reverify identity immediately before launch or uninstall, and tell the user when an action was skipped instead of failing silently.
- Unmounting is never gated: it removes the bind mount and the module directory only, so it stays available even when the underlying package can no longer be identified.
- Reconcile the recorded version with the running one only for a confirmed tracked install.
- Apply the same validated-file and verified-install gates in Settings → Patched APKs.
- Delete the APK stored under the original package name only when no other record owns that path.

### Keeping it cheap

- Certificate extraction verifies the whole archive, so `PM` now caches archive signature hashes keyed by path, size and modification time. The saved APK is parsed once and the resulting `PackageInfo` is carried in the snapshot, so identity, certificates, label and version no longer cost one archive read each.
- Refresh tracked-install state on package add, remove and replace broadcasts, but only for tracked packages and coalesced through a short debounce, so a store updating apps in the background no longer triggers a rebuild per broadcast. The receiver is registered as `RECEIVER_NOT_EXPORTED`.
- Home inspections stay capped by a semaphore, and the duplicate state resolution that ran during version reconciliation was removed.

## Known limitation

An install attributed to the Play Store with no retained patched APK stays `Unverified`: the attribution is the expected one and there is no certificate to compare against. The UI is honest about it and offers the way out, but confirming those installs would mean recording the artifact's certificate at install time, which needs a schema migration and belongs in a separate change. The default configuration retains patched APKs, so this affects users who turned that off.

### DIFF
--- a/.github/workflows/build_pull_request.yml
+++ b/.github/workflows/build_pull_request.yml
@@ -53,6 +53,13 @@ jobs:
       - name: Cache Gradle
         uses: burrunan/gradle-cache-action@v3
 
+      - name: Run unit tests
+        env:
+          _JAVA_OPTIONS: "-Djava.awt.headless=true"
+        run: |
+          chmod +x ./gradlew
+          ./gradlew :app:testDebugUnitTest --stacktrace
+
       - name: Build APK
         env:
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
--- a/app/build.gradle.kts
+++ b/app/build.gradle.kts
@@ -121,7 +121,8 @@ dependencies {
     // Semantic versioning parser
     implementation(libs.semver.parser)
 
-    testImplementation(kotlin("test-junit"))
+    // Unit tests
+    testImplementation(libs.kotlin.test.junit)
 }
 
 android {
--- a/app/build.gradle.kts
+++ b/app/build.gradle.kts
@@ -120,6 +120,8 @@ dependencies {
 
     // Semantic versioning parser
     implementation(libs.semver.parser)
+
+    testImplementation(kotlin("test-junit"))
 }
 
 android {
--- a/app/src/main/java/app/morphe/manager/domain/apk/LocalApkSources.kt
+++ b/app/src/main/java/app/morphe/manager/domain/apk/LocalApkSources.kt
@@ -103,6 +103,22 @@ internal fun resolveTrackedPatchState(
 }
 
 /**
+ * Whether the tracked record can still be removed from the app detail view.
+ *
+ * The record outlives the build it describes, so cleanup has to stay reachable whenever the
+ * patched build is no longer accounted for. Only a confirmed patched install with nothing
+ * retained has nothing to clean up, since the record then describes the app on the device.
+ */
+internal fun canRemoveTrackedRecord(
+    installType: InstallType,
+    patchState: InstalledPatchState?,
+    hasSavedApk: Boolean
+): Boolean =
+    hasSavedApk ||
+            installType == InstallType.SAVED ||
+            patchState != InstalledPatchState.Patched
+
+/**
  * The APKs already on the device that an app could be patched from: the original Morphe kept
  * from a previous run, and the app as the system has it installed.
  */
--- a/app/src/main/java/app/morphe/manager/domain/apk/LocalApkSources.kt
+++ b/app/src/main/java/app/morphe/manager/domain/apk/LocalApkSources.kt
@@ -13,13 +13,19 @@ import app.morphe.manager.data.room.apps.installed.InstalledApp
 import app.morphe.manager.domain.repository.InstalledAppRepository
 import app.morphe.manager.domain.repository.OriginalApkRepository
 import app.morphe.manager.domain.repository.PatchBundleRepository
+import app.morphe.manager.util.AOSP_INSTALLER_PACKAGE
+import app.morphe.manager.util.AOSP_INSTALLER_PACKAGE_LEGACY
 import app.morphe.manager.util.AppDataResolver
 import app.morphe.manager.util.AppDataSource
+import app.morphe.manager.util.PLAY_STORE_INSTALLER_PACKAGE
 import app.morphe.manager.util.PM
 import app.morphe.manager.util.tag
 import kotlinx.coroutines.Dispatchers
 import kotlinx.coroutines.withContext
 import java.io.File
+
+// The record is written once the install finished, so only clock skew separates the two stamps
+private const val INSTALL_TIME_TOLERANCE_MS = 60_000L
 
 /** Saved APK information for display in APK selection dialog. */
 data class SavedApkInfo(
@@ -47,17 +53,31 @@ enum class InstalledPatchState {
     Unknown
 }
 
+/**
+ * Everything a UI action needs to know about a tracked app, resolved in one pass.
+ *
+ * [patchState] is null exactly when nothing is installed under the tracked package name, and
+ * [savedPatchedApkInfo] is the archive parse behind [savedPatchedApk] so callers can read the
+ * label or version without opening the file again.
+ */
 data class TrackedAppSnapshot(
     val installedPackageInfo: PackageInfo?,
     val savedPatchedApk: File?,
+    val savedPatchedApkInfo: PackageInfo?,
     val patchState: InstalledPatchState?
 )
 
+/**
+ * Weighs the evidence that the installed package still is the build Morphe patched.
+ * Kept free of Android APIs so the ordering between the signals can be tested directly.
+ */
 internal fun resolveTrackedPatchState(
     installedHashes: Set<String>,
     savedPatchedHashes: Set<String>,
     originalHashes: Set<String>,
-    installedByPatchManager: Boolean
+    installedByPatchManager: Boolean,
+    installerAttributionMatches: Boolean,
+    installedAfterPatching: Boolean
 ): InstalledPatchState {
     // A mismatch only proves that the installed package is not the reference artifact; it does
     // not identify what replaced it. Only positive matches and trusted installer evidence can
@@ -71,6 +91,14 @@ internal fun resolveTrackedPatchState(
     }
 
     if (installedByPatchManager) return InstalledPatchState.Patched
+
+    // The record knows which installer Morphe would have attributed its own install to. A
+    // different installer on a package that only appeared after the patch is somebody else's
+    // installation, which is the one case certificates cannot describe when nothing was retained.
+    if (!installerAttributionMatches && installedAfterPatching) {
+        return InstalledPatchState.NotPatched
+    }
+
     return InstalledPatchState.Unknown
 }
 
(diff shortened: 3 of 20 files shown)


MarcaDian:
Since your PR is quite complex and deals with critical areas, I ran it through a powerful AI agent; it identified several issues, most notably a performance drop on the main screen caused by the parsing of every application. I’ve attempted to optimize and fix this, as I don't think anyone else will be maintaining this code.

Please test my changes on your end to ensure nothing broke (it shouldn't have, but it's worth verifying). I’ll likely do some testing myself, but please test the specific edge cases you were trying to address, given that you are the author of this PR.

I didn't want to add tests to the project since I’m not very experienced in that area, I haven't worked on projects of this scale before, but sooner or later, it had to be done. I realize that you generated this PR using an AI agent; it’s evident from the code and comments. Please don't hesitate to mention that, there’s nothing wrong with it. However, we do need to know which model was used so we can assess the depth of the code analysis.

AndreaCovelli:
Thanks for taking this on. I reviewed the changes in `718decaa` and `cd3d0b38` against `2f56d42a`, including the cache/snapshot reuse, installer/timestamp evidence, package-broadcast coalescing, action revalidation, and the added test harness.

I found two things I think we should address:

1. **The saved-APK version check has been removed, which reopens one of #830's cases.** `readSavedApkInfo()` deliberately validates only the package and signature, while `validatedPatchedApk()` trusts that the expected filename identifies the version. A valid signed APK for the same package but a different version can therefore be left or renamed to that path and accepted as the saved patched artifact. Its certificate is then used as the strongest `Patched` signal; if that file is a stock APK, a stock installation with the same signer can be classified as patched and regain uninstall/launch actions.

   The stated reason for omitting the check is that patches may rewrite `versionName`, but both the single-app flow ([`PatcherViewModel.kt`](https://github.com/MorpheApp/morphe-manager/blob/cd3d0b38c2dfae6a2439be24726d2fbfc0fefd34/app/src/main/java/app/morphe/manager/ui/viewmodel/PatcherViewModel.kt#L572-L635)) and batch flow ([`BatchPatchCoordinator.kt`](https://github.com/MorpheApp/morphe-manager/blob/cd3d0b38c2dfae6a2439be24726d2fbfc0fefd34/app/src/main/java/app/morphe/manager/domain/batch/BatchPatchCoordinator.kt#L517-L570)) already persist the patched output's `versionName` as `InstalledApp.version` and use it in the retained filename. Comparing the parsed archive version with `app.version` should therefore preserve version-rewriting patches while rejecting the wrong-version artifact documented in #830 and [the proposed #786 follow-up](https://github.com/MorpheApp/morphe-manager/issues/786#issuecomment-5181652473). I suggest restoring that comparison and covering it with a regression test.

2. **The new action-blocked messages are too definitive for `Unknown`.** Both action gates reject every state other than `Patched`, including `Unknown`, but `uninstall_app_unverified` and `launch_app_unverified` say the installed app “is no longer” the tracked patched build. In the unknown case we explicitly do not know that. These should use the same neutral wording as the dialog—e.g. “Morphe could not verify that the installed app is the build it patched”—so the UI does not turn uncertainty into a false stock verdict.

Validation on the current PR tip:

- `./gradlew :app:testDebugUnitTest :app:assembleDebug` with Java 17: **BUILD SUCCESSFUL**
- 45 tests passed: 14 tracked-state, 19 patch-selection, and 12 version tests
- GitHub's `build` check is passing

I have not installed this branch on the device yet, so the Ampere uninstall → Play Store reinstall flow still needs an on-device pass after the points above are resolved.

For transparency, the initial implementation and this re-review were assisted by OpenAI Codex from the GPT-5 family.

MarcaDian:
You're right about the version check, my reasoning for dropping it was just wrong. The single and batch flows are the only two writers of that path and both save the patched output's own `versionName`, so comparing it can't break version-rewriting patches.

Two things worth noting though. That folder is app-private (`getDir` with `MODE_PRIVATE`), so getting a file in there needs root or a bug on our side. And the check narrows the hole rather than closing it: a stock APK with the *same* version at that path still passes and still reads as `Patched`, which is exactly the #830 shape. Only recording the signer at install time really fixes that, which is the schema change under Known limitation.

AndreaCovelli:
I reviewed `78cd1d4f` and `ad52fe7c`. Both findings are resolved:

- The saved-APK predicate now requires the parsed package, the exact version persisted from the patched output, and a non-empty signature set. The new tests cover wrong versions, renamed packages, unsigned/unreadable archives, and a version-rewriting patch whose produced version is also the recorded version.
- The blocked launch/uninstall messages now say Morphe could not verify the build, so they are accurate for both `Unknown` and a confirmed replacement.

I agree with the remaining boundary you described: the exact-version check closes the wrong-version case documented in #830, but it cannot prove that a same-package, same-version file in the private directory was actually produced by Morphe. Persisting artifact signer/hash identity would be the complete solution and would require the separate schema work already called out under Known limitation.

I also checked the intervening `dev` merge. The PR is mergeable, and the rerun completed `:app:testDebugUnitTest` and the APK build successfully.

No further findings in these two commits. The only remaining validation item from my side is the on-device Ampere uninstall → Play Store reinstall flow.

AndreaCovelli:
The model used is specifically GPT 5.6 Sol xhigh

MarcaDian:
So, should we wait for your final check on the device? That would be great.

Since your PR is quite complex, it would be better to merge it before making further changes, and the changes need to be validated.

AndreaCovelli:
Final on-device validation is complete against `ad52fe7c` on a Pixel 9.

I installed the PR as the separate debug variant and tested it against the real Play Store Ampere `v4.36.1` currently on the device. Because the installed release Morphe is non-debuggable, I could not copy its private database, and I did not destructively uninstall the user's Ampere data. Instead, I reproduced the same post-reinstall state by seeding the debug app's private Room database with the stale tracked Ampere row.

Results:

- With a patch timestamp predating the current Play Store installation, Ampere was shown as **Uninstalled**, never **Installed**. Open, uninstall, and install actions were unavailable; only repatching was offered.
- With `patched_at = NULL`, matching how an older `1.25.0` row can migrate without the new evidence, Ampere was shown as **Unverified**, never **Installed**. The detail message remained neutral and the same actions were gated.
- I placed the real signed Ampere `v4.36.1` APK at the exact private saved-APK path for a record claiming `v4.36.0`. It was rejected and the UI reported **No saved APK**, confirming the exact-version validation on-device.
- No crashes, SQLite errors, or invalid actions appeared.

I removed the debug build and all synthetic test data afterward; release Morphe and Play Store Ampere were left unchanged.

So the final device check passes, including the conservative migration fallback.

AndreaCovelli:
I'm doing some review of the changes from a bird's-eye view, I'd like this pr of mine to be robust

MarcaDian:
If related issues arise, it is better to handle them in subsequent PRs rather than lumping everything together.

AndreaCovelli:
Yeah, I agree, I'll come back to you when I think it's ready

AndreaCovelli:
One scope check before I consider the review complete: of the two UI observations from the device test, I think only one plausibly belongs in this PR.

When a tracked install is `Unknown` and there is no retained patched APK, the detail view offers Patch but no way to delete/forget the retained tracking row. Exposing the existing record-cleanup action for that state would leave the currently installed package untouched; it would just let the user remove Morphe's stale entry. Since `Unknown` and its action gating are introduced here, would you prefer that minimal fix (with a focused regression test) in #831, or should I file it for a follow-up too?

I consider the separate wording issue—an installed `NotPatched` package currently being labelled `Uninstalled`—a presentation refinement and can leave that entirely to a follow-up. I won't change the branch until you choose.

MarcaDian:
Since this PR is very complex and is holding up subsequent changes, I’m going to merge it now to avoid errors during future merges. If you have any further edits, please open a new PR.

PR #849, opened 2026-08-11
feat: Add CPU and storage I/O graphs to the expert patching screen

### LINKED ISSUE #624: enhance: Place "Games" button at the bottom
### Feature description

Hi! I'd like to request the "Games" button to be placed at the bottom.

|Image|Image|
|---|---|

What did I change?
- "Games" button is now placed at the bottom (of course)
- "Games" simplified to an icon to fit the button
- "Copy" button moved to the Patcher Logs section itself
- "Patcher Logs" button reverted to the previous look (from v1.18.0)
- Games UI slightly taller as it does not need a "Patcher Logs" header anymore

### Motivation

This enhancement request allows for two things: 

1. **Games can now be accessed in Simple Mode.** 

Now, I know everyone reading this right now has Morphe set to Expert Mode, but there are still a handful of users in Simple Mode. I believe (and I hope you too) that those people deserve access to Morphe's built-in games. 

By putting the "Games" button at the bottom, there is now a way to fit the button to the Simple Mode UI.

|Image|Image|
|---|---|

2. **Wasted space can now be utilized in two-panel mode (tablets).**

In two-panel mode, there's wasted space in the left side, as visualized below:

|Image|Image|
|---|---|

This change fixes that. It takes advantage of the space to put the games section in its dedicated spot. Now you can view your patch logs while playing 2048 (I wouldn't recommend looking on other games though 🫠)! 

|Image|Image|
|---|---|

<details>
<summary>P.S. another concept :P</summary>

If the new copy button looks weird, I made another version where the button is expanded and has the word "Copy" in it:

|Image|Image|
|---|---|

I personally *don't prefer it* over the original design, but it's an option regardless.

</details>

### Acknowledgements

MarcaDian:
### Description:
Expert mode now shows what a patch run is actually doing to the device, not just how much heap it holds.

**New graphs**
- **CPU** - load of every core, sampled live. On devices that let apps read system-wide CPU stats these are the real per-core numbers; elsewhere they are derived from the patcher's own threads, attributed to the core each one last ran on.
- **Storage I/O** - read and write throughput of the patcher process, with the current rate split into ↓ read and ↑ write.

**Redesigned memory graph**
All three panels share the same bar rendering, with a colored accent per panel and a larger current reading. Memory and CPU shift towards the error color as they fill up.

**Layout**
- Phone: three compact panels share a single row.
- Tablet and landscape: each panel gets its own row and fills the previously empty left column, and that column now scrolls so nothing is cut off on short screens.
- Panel sizes follow the window, so a phone on its side gets a tighter layout than a tablet.

Sampling adds no measurable cost to patching: readings are taken once every two seconds on a background-priority thread, and a metric the device does not expose is simply left out rather than shown as zero.
___


dr-strange15:
Hii

bar graphs Will look better than line graphs for Memory Usage and Storage I/O, making the metrics easier to glance at.

Would also make the overall UI feel more consistent with the CPU usage graph😅

PR #851, opened 2026-08-11
fix: Distinguish present non-patched tracked apps

### LINKED ISSUE #850: bug: Confirmed non-patched tracked app is shown as uninstalled
### Bug description

After #831, Morphe can positively identify that a package is present but is not the tracked patched build. The UI currently maps that `NotPatched` state to the same `isDeleted` presentation used when no package is installed.

#### Steps to reproduce

1. Patch an app with Morphe and install the patched build.
2. Uninstall it outside Morphe.
3. Install the stock build again under the same package name.
4. Open Morphe after it identifies the current package as `NotPatched`.

#### Actual behavior

The home card displays **Uninstalled**, and the detail dialog says **App was uninstalled**, even though an app with that package name is installed on the device.

#### Expected behavior

A confirmed present `NotPatched` package should use the existing **Not patched yet** presentation. The absent-package state should remain **Uninstalled**, and the `Unknown` state should remain **Unverified**. Existing action gating and tracked-record cleanup behavior should not change.

This is a presentation follow-up to #831. During that PR's Pixel 9 validation, the Play Store Ampere reinstall was correctly classified as not being Morphe's tracked patched build, but it exposed this inaccurate label.

- Morphe Manager `dev` after #831 (`v1.26.0-dev.3`)
- Ampere v4.36.1 (`com.gombosdev.ampere`)
- Pixel 9

### Morphe logs

```shell
Not applicable: this is a deterministic presentation-state mismatch and no crash or runtime error occurs.
```

### Acknowledgements

AndreaCovelli:
## Summary

- present a confirmed non-patched replacement as **Unpatched version installed**, distinct from an app that Morphe has never patched
- show the replacement's current installed icon and version and classify it as installed, while retaining Morphe's previous patch record and cleanup action
- reserve **Uninstalled** for an absent package and keep `Unknown` conservative as **Unverified**

Fixes #850. Follow-up to #831.

## Root cause

#831 introduced distinct `Patched`, `NotPatched`, and `Unknown` evidence states, but the presentation layer mapped both `NotPatched` and an absent package to `isDeleted`. It also displayed retained patched-APK metadata for both `NotPatched` and `Unknown`. A confirmed Play Store or stock replacement was therefore presented as an uninstalled historical build rather than the different, unpatched package actually on the device.

## Changes

- Centralize the tracked-state presentation mapping used by the home screen and detail dialog.
- Treat `NotPatched` as a present confirmed replacement, use its current PackageManager metadata, and label it **Unpatched version installed**.
- Keep its historical patched record and detail screen, including Patch and Delete, while continuing to withhold Open and Uninstall actions.
- Keep `Unknown` on retained metadata and prevent the normal untracked-app resolver from leaking the current package identity when no retained APK exists.
- Normalize displayed version prefixes so an installed version already beginning with `v` is not rendered with a duplicate prefix.
- Cover the presentation, metadata selection, filter membership, and version formatting contracts with focused unit tests.

## Validation

- `:app:testDebugUnitTest :app:assembleDebug`: **BUILD SUCCESSFUL**
- 74 unit tests passed
- Pixel 9 device test against real Play Store Ampere `v4.36.1`:
  - confirmed replacement showed its current icon/version and **Unpatched version installed**, never **Uninstalled**
  - detail showed the previous-record explanation with Patch and Delete only
  - adjacent `Unknown` case showed retained `v4.36.0` as **Unverified** and did not reveal current package metadata
  - no AndroidRuntime, Room, or SQLite errors
- Debug app and synthetic device artifacts removed afterward; release Morphe and Ampere remained unchanged
- `git diff --check`: clean

### DIFF
--- a/app/src/main/java/app/morphe/manager/ui/model/HomeAppItem.kt
+++ b/app/src/main/java/app/morphe/manager/ui/model/HomeAppItem.kt
@@ -20,6 +20,7 @@ data class HomeAppItem(
     val isPinnedByDefault: Boolean,
     val isInstalledOnDevice: Boolean,
     val isDeleted: Boolean,
+    val isInstallStateNotPatched: Boolean,
     val isInstallStateUnknown: Boolean,
     val savedApkFile: File?,
     val hasUpdate: Boolean,
@@ -31,5 +32,8 @@ data class HomeAppItem(
      * Whether the pending update is worth surfacing in the UI.
      * Uninstalled apps keep their update flag but show the uninstalled state instead.
      */
-    val showsUpdateBadge: Boolean get() = hasUpdate && !isDeleted && !isInstallStateUnknown
+    val showsUpdateBadge: Boolean get() = hasUpdate &&
+            !isDeleted &&
+            !isInstallStateNotPatched &&
+            !isInstallStateUnknown
 }
--- a/app/src/main/java/app/morphe/manager/ui/model/TrackedInstallPresentation.kt
+++ b/app/src/main/java/app/morphe/manager/ui/model/TrackedInstallPresentation.kt
@@ -1,0 +1,37 @@
+/*
+ * Copyright 2026 Morphe.
+ * https://github.com/MorpheApp/morphe-manager
+ */
+
+package app.morphe.manager.ui.model
+
+import app.morphe.manager.data.room.apps.installed.InstallType
+import app.morphe.manager.domain.apk.InstalledPatchState
+
+/** User-facing state derived from Morphe's evidence about a tracked install. */
+internal data class TrackedInstallPresentation(
+    val isPatched: Boolean = false,
+    val isDeleted: Boolean = false,
+    val isNotPatched: Boolean = false,
+    val isUnknown: Boolean = false
+)
+
+/**
+ * Keeps a present, confirmed non-patched package distinct from a package that is absent.
+ *
+ * [isDeleted] still accounts for the tracked patched build being gone so existing recovery and
+ * filtering behavior stays unchanged. [isNotPatched] lets the UI describe the package that is
+ * actually present instead of claiming that no app is installed.
+ */
+internal fun trackedInstallPresentation(
+    installType: InstallType,
+    patchState: InstalledPatchState?
+): TrackedInstallPresentation = when (patchState) {
+    InstalledPatchState.Patched -> TrackedInstallPresentation(isPatched = true)
+    InstalledPatchState.NotPatched -> TrackedInstallPresentation(
+        isDeleted = installType != InstallType.SAVED,
+        isNotPatched = true
+    )
+    InstalledPatchState.Unknown -> TrackedInstallPresentation(isUnknown = true)
+    null -> TrackedInstallPresentation(isDeleted = installType != InstallType.SAVED)
+}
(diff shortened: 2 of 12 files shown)


AndreaCovelli:
On-device validation is complete for `d55cab66` on a Pixel 9.

I installed the PR as the parallel debug variant and tested it against the real Play Store Ampere `v4.36.1` (`com.gombosdev.ampere`, installed by `com.android.vending`). Only the debug app's private Room database was seeded with a synthetic tracked `v4.36.0` record.

Results:

- A confirmed replacement showed Ampere's current icon and `v4.36.1` with **Unpatched version installed** on home and in the detail hero. It was not presented as **Uninstalled** or as an app Morphe had never patched.
- The detail explanation preserved the previous-record context. Patch and Delete were available; Open, Install, and Uninstall were not.
- The adjacent `Unknown` case remained **Unverified**, displayed retained `v4.36.0` rather than current package metadata, and kept Patch/Delete only.
- Device testing exposed and verified fixes for duplicate `v` prefixes and the Unknown-state fallback to normal installed metadata.
- No AndroidRuntime, Room, or SQLite errors occurred.

The debug package, synthetic database, and temporary device artifacts were removed afterward. Release Morphe and Play Store Ampere remain unchanged. Local validation passes with 74 unit tests and a successful debug APK build.

AndreaCovelli:
@marcadian One presentation question came up during the device test, and I would appreciate your preference before changing this further.

For a confirmed `NotPatched` replacement, the PR currently shows the package actually installed on the device (Ampere `v4.36.1`). Morphe has positively identified that package as an ordinary/unpatched replacement, so showing the old tracked `v4.36.0` alone could imply that the previous patched build is still installed.

For `Unknown`, it instead shows the retained record (`v4.36.0`), because Morphe cannot safely associate the current package metadata with its tracked build.

The most explicit model would be:

- **Installed now:** `v4.36.1` — unpatched
- **Previous Morphe record:** `v4.36.0`

For `Unknown`, the first line would remain unverified while the previous record stays visible.

My preference is to keep the current installed version on the home card and, if the historical version needs to be surfaced, show both with explicit labels in the detail view. Do you prefer the current behavior for this PR, or would you want the dual-version detail added here rather than in a small follow-up PR?

MarcaDian:
These AI comments are starting to get tiresome. Can we communicate without them?
I don't understand who I am communicating with, you or your model.

rushiranpise:
After your PR #831, I’ve noticed a significant delay when loading the app list. I’m not sure if this is device-specific, but I do have 200+ apps. Previously, the app list loaded fine within a few seconds.

I’ve attached a video. I can provide logs if necessary.

https://photos.app.goo.gl/GXX1sF8K3fwLLSdt8

AndreaCovelli:
> These AI comments are starting to get tiresome. Can we communicate without them? I don't understand who I am communicating with, you or your model.

@MarcaDian you're right, I'll try to write the next ones myself.

AndreaCovelli:
Regarding the UI, do you agree with showing the currently installed version for a confirmed unpatched replacement, while an unverified app shows the previous Morphe record? Or would you prefer the detail screen to show both the currently installed version and the previous patched version? I think showing the current version on the card and both versions in the details would be clearest.

MarcaDian:
At the moment, I think the current appearance is acceptable.

MarcaDian:
It is more important to fix the performance loss. I have already prepared a fix #852 due to database usage, but now it needs testing. I am waiting for confirmation from  @rushiranpise

AndreaCovelli:
Yeah, I've seen, thanks for your work. I'll also test it on my pixel 9.

PR #852, opened 2026-08-12
fix: Stop re-verifying every tracked APK on each cold start

MarcaDian:
### Description

Since #831 the home screen resolved a `TrackedAppSnapshot` for every tracked app on each refresh, and each snapshot ran up to three full archive verifications: the saved patched APK, the installed `base.apk`, and the saved original APK. `getPackageArchiveInfo` with `GET_SIGNING_CERTIFICATES` hashes the entire file, so a library of large apps meant hashing around a gigabyte on every cold start, with the home state blocked until it finished. Users with many tracked apps reported 40+ seconds before any card appeared.

A simpleperf trace of the first 30 seconds after a cold start put `sha512_block_data_order` at 7.7% of all cycles, 155.9 billion cycles total.

### Changes

- Certificate fingerprints are persisted in a new `apk_signatures` table (schema 12 → 13), keyed by path, size and mtime, so a rewritten file is never answered from a stale entry.
- `readSavedApkInfo` decides identity from the manifest alone and asks for certificates only once it matches, so an archive that is not the recorded artifact is rejected without hashing.
- The mount check (`hasSourceApkSignatureMismatch`) is read only when it could still change the verdict, rather than ahead of every other signal. The resulting state table is unchanged.
- `TrackedAppSnapshot` is cached per package behind a fingerprint built from stat calls, and invalidated by the package-change receiver and by `notifyAppStateChanged`.
- Snapshots resolve in their own flow instead of inside the home state. Cards render from PackageManager and adopt the verified state as it lands, so verification never holds them back.

### DIFF
--- a/app/schemas/app.morphe.manager.data.room.AppDatabase/13.json
+++ b/app/schemas/app.morphe.manager.data.room.AppDatabase/13.json
@@ -1,0 +1,521 @@
+{
+  "formatVersion": 1,
+  "database": {
+    "version": 13,
+    "identityHash": "d8524a5875f279ef0e9e1cd214583b55",
+    "entities": [
+      {
+        "tableName": "patch_bundles",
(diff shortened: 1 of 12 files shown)


AndreaCovelli:
Ran some testing on a Pixel 9 (v1.26.0-dev.5 / c2bb6598). I got these results:

Workload: 10 tracked apps with real APKs (559.6 MiB total)
Uncached cold launch: ~1.84s, cards popped up right away on initial load
Cached cold launch: Settled around 1.60s median (1.59s-1.67s range). Cache correctly held onto all 10 APK signatures
Migration: Schema 12 → 13 ran cleanly in 1.65s without breaking DB integrity

No crashes, ANRs, or database/cache errors popped up throughout testing.

MarcaDian:
Great.

PR #853, opened 2026-08-12
fix: Keep tracked install verification state fresh

AndreaCovelli:
### Description

Follow-up to #852. Snapshot verification no longer blocks the initial app list, but unresolved records could briefly reuse an old verdict or appear as installed before their result arrived. The snapshot fingerprint also omitted saved-original APKs and bundle certificate changes.

### Changes

- Publish tracked snapshots per app and show `Checking install…` until each verdict arrives.
- Remove an affected verdict as soon as a package change is reported, while keeping archive work debounced.
- Re-evaluate snapshots when saved-original APKs or bundle signatures change, and include all verdict inputs in the snapshot fingerprint.
- Bind signature-cache writes to the file revision inspected and prevent stale cleanup or delayed writes from replacing a newer cache row.
- Add focused tests for pending presentation, evidence invalidation, and APK revision stamps.

### Validation

- `./gradlew :app:testDebugUnitTest` — 79 tests passed.
- `./gradlew :app:assembleDebug` — passed.
- Pixel 9: tested five real tracked APKs totaling 1,067.9 MiB with an empty cache; all five resolved correctly, five cache rows were persisted, the database remained valid, and no crash/ANR/cache errors were logged.

### DIFF
--- a/app/src/main/java/app/morphe/manager/data/room/apk/ApkSignatureDao.kt
+++ b/app/src/main/java/app/morphe/manager/data/room/apk/ApkSignatureDao.kt
@@ -14,6 +14,13 @@ interface ApkSignatureDao {
     @Insert(onConflict = OnConflictStrategy.REPLACE)
     suspend fun upsert(signature: ApkSignature)
 
-    @Query("DELETE FROM apk_signatures WHERE file_path IN (:filePaths)")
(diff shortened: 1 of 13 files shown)


AndreaCovelli:
@MarcaDian I (me and codex) did a review of your changes and found those medium findings.

Issue #103, opened 2026-01-11
feat: Downloaders Support


decipher3114:
### Feature description

Allow using Downloaders for Manager (in both modes)

### Motivation

Using Downloader will make it much easier tom download and manage apk files.

### Acknowledgements

- [x] I have checked all open and closed feature requests and this is not a duplicate.
- [x] I have chosen an appropriate title.
- [x] The feature request is only related to Morphe (Manager).

EithonX:
+1

optimatrix0:
In my view, I would prefer an older approach where the user can choose which browser to use. This allows selecting a preferred browser, using ad blockers and tracking protection, and having more flexibility overall. I think the Downloader approach is valid, but in general, using a browser offers a much better experience.

So here is my honest suggestion: it might be worth considering an option to open links with a browser like Firefox, which supports offline translation and has full support for ad blockers. That could be a great alternative.

Thanks to the author of this post for bringing this issue up. It’s important to surface these points so we can discuss and collaborate on improvements. The Downloader proposal makes sense, but in my specific use case, giving users the choice of which browser to open with would be even better.

DraconicNEO:
I really think it would be good to have downloader support for loading the APKs directly from APKmirror instead of prompting a browser or google search. Especially since they chose to remove advanced mode features and therefore the ability to use downloaders at all which is pretty shitty in my opinion.

akshaynexus:
im looking at porting apkeep to this repo to make this easier,however this uses APKPure as source ,apkmirror doesnt have a easy cli/tool to download the apks

Another option is to use google play store as source but it needs AAS Token or ya29 auth token which is bad UX

so mainly commentint in this  issue to see if this would be worth implementing or better find a better ux to use the prot of aptool with google play store as source

i think the ideal way is to use apkeep port and use gogole play store as source and use the migrog auth for getting the apk? might be the ideal?

LisoUseInAIKyrios:
Any kind of built in downloader has issues of providing apks in app, and all the apk sites have terms of service that do not allow automation.

Just so it's clear, there are no plans to ever add this and any pull requests to add this function will be closed.

PR #855, opened 2026-08-12
Improved Base APK Downloading and Crash Report System

Lawtro37:
The base APK is now downloaded via a webview which automaticly selects the downloaded APK file and has a fancy dowload screen.
I added this because I thought it would streamline the process a little. I've always hated how APKs need to be retreived from APK mirror so this makes that sigtly cleaner.
For some reason the last part of the download takes longer than the rest doing it this way but it isn't too much of an issue.

When I was making this I also added a crash report system to quickly tell me what was going wrong when the app crashed.
It just opens a new screen with the stack trace whenever an uncaught exception is thrown that can be scrolled through, copied and shared for debugging.

To test this I also added a "test crash" button to the bottom of the system settings screen on expert mode.


LisoUseInAIKyrios:
We don't want to take on the risk of our apps "providing apks".

See comments here
https://github.com/MorpheApp/morphe-manager/issues/103#issuecomment-4723866548

There is now support for external apps to download and provide apks, and it works even better than using any in-app browser and the third party app takes on any risks itself.

Lawtro37:
> We don't want to take on the risk of our apps "providing apks".
> 
> See comments here [#103 (comment)](https://github.com/MorpheApp/morphe-manager/issues/103#issuecomment-4723866548)
> 
> There is now support for external apps to download and provide apks, and it works even better than using any in-app browser and the third party app takes on any risks itself.

This is not an automatic downloader/app provider. It simply adds a webview of Apkmirror and just hooks do the download once the user starts it and shows a fancy little GUI. This also has the bonus of the user not needing to pick the APK.

see attached video
https://discord-host.lawtrostudios.com/v/2a5f7b6f-2f60-441b-9e12-e168f9693e09

LisoUseInAIKyrios:
It's still providing the apk inside the app.

It's not something we want to take on, nor do we need to take on.

Lawtro37:
> It's still providing the apk inside the app.
> 
> It's not something we want to take on, nor do we need to take on.

Sorry for nagging but may I ask why?
I dont see how this system breaks the TOS of Apkmirror or takes on any risk. 
All ads are shown and the user is genuenly browsing the page and the apk is still downloaded through the webview browser.
It doesnt extract the download link and download it in the app that way if thats what your thinking it does.
It simply detects when the webview starts a download and shows you its progress and sellecting that file to be patched.
It's essentualy the same thing your already doing but it automaticly detects the downloaded apk and has a fancy little  dowload screen.
I understand if you dont want to merge my PR, your the maintainer its your choice, I just dont understand your reasoning thats all.

LisoUseInAIKyrios:
It's not a terms of service issue with Apkmirror, it's because this change fundamentally changes who is providing the apk.

If we provide downloading _inside_ our app, then bad actors can claim "Morphe provides copyrighted apks".

PR #859, opened 2026-08-13
fix: Preserve tracked state after deleting saved APKs

AndreaCovelli:
### Description

Follow-up to #853. Deleting an item from Settings > Storage management > Patched APKs removed both the retained file and its tracking row, even though the action and confirmation described file-only cleanup. An installed app therefore fell back to `Not patched yet` instead of being re-evaluated as a tracked install. If that re-evaluation produced `Unverified` with no retained archive, the detail header also had no package information from which to render an icon.

### Changes

- Separate retained-APK deletion from the explicit operation that forgets an app record.
- Keep single, selected, and bulk patched-APK cleanup file-only and refresh the storage list immediately.
- Notify Home when retained verification evidence changes so the old verdict becomes pending and is recomputed.
- Use the inspected installed package only as the detail icon fallback; the tracked label and version presentation remain unchanged.
- Add a focused regression test for the unverified icon/metadata split.

### Validation

- `./gradlew :app:testDebugUnitTest` — 115 tests passed.
- `./gradlew :app:assembleDebug` — passed.
- Pixel 9, isolated debug app: a tracked installed app initially showed `Installed`; deleting its retained APK through the storage UI removed the file and refreshed the list, Home changed to `Unverified`, the tracking row remained present with database integrity `ok`, and the detail header showed the real app icon with the tracked version.
- The isolated debug app and test fixtures were removed afterward; production Morphe and Ampere were untouched.

### DIFF
--- a/app/src/main/java/app/morphe/manager/domain/repository/InstalledAppRepository.kt
+++ b/app/src/main/java/app/morphe/manager/domain/repository/InstalledAppRepository.kt
@@ -157,18 +157,14 @@ class InstalledAppRepository(
 
     /**
      * Update only the [InstalledApp.version] of an existing record and drop the orphan
-     * patched APK at the old version path. Applied patches and selectionPayload are
-     * left untouched.
+     * patched APKs it owned at the old version path. Applied patches and selectionPayload
+     * are left untouched.
      */
     suspend fun updateInstalledVersion(app: InstalledApp, newVersion: String) =
         withContext(Dispatchers.IO) {
             if (app.version == newVersion) return@withContext
             dao.upsertApp(app.copy(version = newVersion))
-            val orphans = listOf(
-                filesystem.getPatchedAppFile(app.currentPackageName, app.version),
-                filesystem.getPatchedAppFile(app.originalPackageName, app.version)
-            ).distinctBy { it.absolutePath }
-            orphans.forEach { file ->
+            savedPatchedApkFiles(app).forEach { file ->
                 if (file.exists()) {
                     runCatching { file.delete() }.onFailure {
                         Log.w(TAG, "Failed to delete ${file.absolutePath}", it)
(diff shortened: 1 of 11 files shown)


AndreaCovelli:
Thanks for the simplification on your end with the commit, AI overengineers a lot

AndreaCovelli:
If I may ask, which AI model do you use?

MarcaDian:
Opus 5, The main thing is to have a proper understanding of what needs to be done.

AndreaCovelli:
Doing a review pass, the model found these which I do think they are correct but minor findings:
1. [Medium] Failed APK deletion leaves behind ghost/orphaned files

In deleteSavedPatchedApkFiles (https://github.com/MorpheApp/morphe-manager/blob/287d7b2c673e7e2cc6c685be9b9a6c671142c021/app/src/main/java/app/morphe/manager/domain/repository/InstalledAppRepository.kt#L188-L218), the method currently returns true if a target file existed, even if File.delete() actually fails to delete it. Because it reports success, the caller goes ahead and wipes the database entry. This leaves the un-deleted APK sitting on storage while completely hiding it from Morphe's UI, all while displaying a misleading "success" toast to the user.

- **Suggested fix:** Keep the SAVED database record intact whenever any file fails to delete, and pass meaningful status back so the UI can notify the user if deletion actually failed.

---

2. [Low] Interrupted APK save can accidentally wipe existing archives

Inside saveOriginalApk (https://github.com/MorpheApp/morphe-manager/blob/287d7b2c673e7e2cc6c685be9b9a6c671142c021/app/src/main/java/app/morphe/manager/domain/repository/OriginalApkRepository.kt#L44-L82), the copies variable only checks whether sourceFile != targetFile—it doesn't actually confirm if this specific operation created the file. If an error occurs right before or during the copy process, execution falls into the catch block and can delete an already-existing target file that was there prior to the save attempt.

- **Suggested fix:** Use a temporary file and swap it atomically upon completion, or at least check if the file existed beforehand and track copy progress before attempting cleanup.

MarcaDian:
Should be fixed in https://github.com/MorpheApp/morphe-manager/commit/f896574c948aadd0eceb3165fab842e2c54319da

PR #868, opened 2026-08-15
fix: Honor a non default port in remote bundle URLs

virzak:
## Problem

A remote patch source whose URL has an explicit port silently loses it.

`PatchBundleRepository.normalizeRemoteBundleUrl` rebuilds the URL for arbitrary hosts as:

```kotlin
return "https://$host$normalizedPath$query"
```

`Url.host` does not include the port, so

```
https://example.com:57158/patches-bundle.json
```

is requested as

```
https://example.com/patches-bundle.json
```

## Why it is awkward to diagnose

Nothing in the UI points at the cause:

- the URL passes validation and the source is added ("2 sources installed");
- the source then sits as **Unnamed** with no error shown;
- the server never receives a request, so it looks like Manager is not fetching at all.

It surfaces only in the log, as a timeout against a URL the user never typed:

```
HttpService.request: Connecting to URL: https://example.com/patches-bundle.json
request failed after 3 attempts: ConnectTimeoutException ...
Caused by: java.net.SocketTimeoutException: failed to connect to example.com/… (port 443)
Failed to update bundle Unnamed (HTTP null)
```

## Change

Append the port when it differs from the protocol default, and use that when
rebuilding the URL for arbitrary hosts.

Deliberately limited in scope:

- only the arbitrary host branch is touched. The GitHub and GitLab branches rewrite to
  fixed hosts (`raw.githubusercontent.com`, `gitlab.com`) where a port is never
  meaningful, so they are left alone;
- URLs on a default port are unaffected, since the port is only appended when it
  differs from `protocol.defaultPort`;
- the scheme is still normalised to https, unchanged from before.

## Verification

Built and installed on a device, against a bundle served over HTTPS on port 57158.

Before, the server receives nothing and the source stays "Unnamed". After, both
requests arrive on the correct port and the bundle loads with its version displayed:

```
"GET /patches-bundle.json HTTP/2.0" 200 244    "Morphe-Manager/…"
"GET /rutube-patches.mpp   HTTP/2.0" 200 657956 "Morphe-Manager/…"
```

I did not add a test: the module has no `src/test` source set, and adding test
infrastructure seemed out of scope for this fix. Happy to add one if you would like it,
`normalizeRemoteBundleUrl` is straightforward to cover.

### DIFF
--- a/app/src/main/java/app/morphe/manager/domain/repository/PatchBundleRepository.kt
+++ b/app/src/main/java/app/morphe/manager/domain/repository/PatchBundleRepository.kt
@@ -26,6 +26,7 @@ import app.morphe.manager.ui.viewmodel.BundleSnapshot
 import app.morphe.manager.util.*
 import io.ktor.client.plugins.ResponseException
 import io.ktor.http.Url
+import io.ktor.http.hostWithPortIfSpecified
 import kotlinx.collections.immutable.PersistentMap
 import kotlinx.collections.immutable.mutate
 import kotlinx.collections.immutable.persistentMapOf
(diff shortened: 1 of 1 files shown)


MarcaDian:
Thanks 👍

PR #879, opened 2026-08-19
fix: Dialog status bar stays all white in light theme

clearpath-mind:
## Problem

**Before**

<div>
  Dialog status bar stays all white in light theme
  Dialog status bar stays all white in light theme (2)
</div>

**After**

Dialog status bar after the fix

`AppDialog` (app details, confirmations, pickers) runs in its own dialog window. That window's status/navigation bar icon appearance was never set, so the icons kept following the app theme's default and stayed **white even in light theme**, where the dialog background is light — making them nearly invisible.

## Fix

Mirror the activity-level behavior in `Theme.kt:156-157` for the dialog window:

```kotlin
WindowCompat.getInsetsController(it, it.decorView).isAppearanceLightStatusBars = !isDarkTheme
WindowCompat.getInsetsController(it, it.decorView).isAppearanceLightNavigationBars = !isDarkTheme
```

Added inside the existing `SideEffect` in `AppDialog.kt` where the dialog window is already configured (dim, transparent background). Icons are now black in light theme and white in dark theme, matching the activity.

## Test plan

- Light theme: open a dialog (e.g. app details) → status/nav bar icons are black
- Dark theme: same dialogs → icons stay white

### DIFF
--- a/app/src/main/java/app/morphe/manager/ui/screen/shared/AppDialog.kt
+++ b/app/src/main/java/app/morphe/manager/ui/screen/shared/AppDialog.kt
@@ -132,16 +132,7 @@ fun AppDialog(
             decorFitsSystemWindows = false
         )
     ) {
-        // Remove standard system backgrounds/window shadows
-        val dialogWindow = (LocalView.current.parent as? DialogWindowProvider)?.window
-        SideEffect {
-            dialogWindow?.let {
-                it.setDimAmount(0f)
-                it.setBackgroundDrawableResource(android.R.color.transparent)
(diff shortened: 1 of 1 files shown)


MarcaDian:
Which Android version do you have? <s>Because I can't reproduce this bug.</s>

Actually, to reproduce it, need to set the light theme on the device. Yes, the problem exists.

clearpath-mind:
I'm on a Xiaomi Redmi Note 10, Android 13

MarcaDian:
Thanks 👍

PR #885, opened 2026-08-20
Optimize manager startup and reduce repository initialization overhead

jo2dan94:
## Summary

This PR contains a set of startup and repository-performance optimizations developed through repeated profiling, controlled experiments, and regression testing of Morphe Manager's cold-start path.

The primary goal was to reduce work performed on or competing with the critical startup path without changing application behavior.

The final optimized build reduced the measured Activity → Home-first startup interval from approximately:

| Metric | Baseline | Optimized | Difference |
|---|---:|---:|---:|
| Activity → Home first | ~698 ms | ~548 ms | **-150 ms (-21.5%)** |
| Activity → first frame | ~49 ms | ~44 ms | **-5 ms** |
| Source build | — | ~49 ms | — |
| Metadata | ~945 ms | ~851 ms | **-94 ms** |
| Reload | ~1091 ms | ~974 ms | **-117 ms** |

The relatively small change in Activity → first-frame time is important: the startup improvement primarily comes from removing/defering expensive application work from the period between Activity creation and Home becoming usable, rather than simply changing when the first frame is reported.

The submitted branch contains only the production optimizations.

All profiling instrumentation, timing markers, experimental code, and unsuccessful optimization attempts used during development have been removed.

---

# Optimization details

### Files

- `ManagerApplication.kt`
- `HomeScreen.kt`
- `HomeViewModel.kt`
- `ViewModelModule.kt`

### Problem

`PatchBundleRepository` is one of the more expensive dependency graphs initialized during startup.

Profiling showed repository-related work beginning before or around Activity startup and therefore competing with initial Home composition.

Repository construction can involve:

- source construction
- bundle metadata
- installed bundle inspection
- preferences
- patch bundle loading
- repository flows
- database/cache access
- source synchronization

Much of this work is required by the application, but it does not all need to block the first usable Home UI.

### Change

Repository initialization is now coordinated with the Home startup lifecycle.

Home signals when its first frame has been reached, allowing repository initialization that does not need to block initial UI construction to begin afterward.

Consumers that require the repository continue to wait for it normally.

This preserves repository semantics while preventing expensive repository construction from unnecessarily competing with initial Home rendering.

### Why this is safe

This does not:

- change repository singleton semantics
- remove repository initialization
- skip repository reloads
- change repository data
- change source ordering
- change update behavior

It changes **when expensive initialization is allowed to enter the startup path**, not what the repository eventually does.

### Profiling evidence

Earlier attribution showed repository-backed startup operations beginning around Activity startup.

For example, InstalledAppRepository access was observed at approximately:

    InstalledAppRepository access start: -40 ms relative to Activity
    InstalledAppRepository access elapsed: ~303 ms

Provider attribution showed:

    InstalledAppRepository provider: ~299 ms
    PatchBundleRepository provider: ~287 ms

This demonstrated that substantial dependency/repository construction was happening concurrently with the initial UI startup period.

However, controlled follow-up experiments also demonstrated that indiscriminately deferring individual repository consumers was not beneficial.

Examples:

    Shortcut observer-only deferral:
        Activity → Home first: 582 ms
        P40 keeper:             548 ms
        Regression:             +34 ms

    Avatar/PatchBundleRepository.sources-only deferral:
        Activity → Home first: 562 ms
        P40 keeper:             548 ms
        Regression:             +14 ms

    Blocklist-only deferral:
        Activity → Home first: 586 ms
        P40 keeper:             548 ms
        Regression:             +38 ms

    Combined three-path deferral:
        Activity → Home first: 597 ms
        P40 keeper:             548 ms
        Regression:             +49 ms

Those negative experiments were intentionally discarded.

The implementation in this PR therefore retains only the startup boundary that survived controlled testing rather than broadly delaying repository consumers.

---

# 2. Lazy PatchBundleRepository resolution in HomeViewModel

### Files

- `HomeViewModel.kt`
- `ViewModelModule.kt`

### Problem

Constructing `HomeViewModel` could transitively force construction of `PatchBundleRepository`.

That meant creating the ViewModel itself could trigger a comparatively expensive repository dependency graph even when initial Home composition did not immediately require every repository-backed value.

### Change

`PatchBundleRepository` is supplied to `HomeViewModel` as a synchronized lazy dependency.

Repository-backed state/flows are also initialized lazily where appropriate.

Conceptually, startup changes from:

    HomeViewModel creation
        ↓
    immediately resolve PatchBundleRepository
        ↓
    construct repository dependency graph
        ↓
    continue Home startup

to:

    HomeViewModel creation
        ↓
    create lazy PatchBundleRepository reference
        ↓
    continue Home startup
        ↓
    resolve repository when actually required

### Benefit

This breaks an unnecessary eager dependency edge in the startup graph.

It allows ViewModel creation and initial Home work to proceed without automatically forcing PatchBundleRepository construction.

### Timing attribution

Repository construction was measured in the ~287–299 ms range during provider attribution.

That does **not** mean this optimization saves 287–299 ms of wall-clock startup time—the repository still has to be constructed and some of that work overlaps other startup work.

The measurable benefit is instead part of the overall ~150 ms Activity → Home-first improvement.

---

# 3. Lazy batch-patching dependencies in MainViewModel

### File

- `MainViewModel.kt`

### Problem

`MainViewModel` previously received batch-patching components eagerly:

- `BatchPlanResolver`
- `BatchPatchCoordinator`

These components are needed for batch operations but are not required merely to display the initial application UI.

Eager constructor injection therefore allowed batch-related dependency construction to become part of cold startup.

### Change

Both dependencies are now resolved lazily only when batch functionality is actually used.

### Result

Cold startup no longer needs to construct unrelated batch-patching infrastructure simply because `MainViewModel` exists.

### Behavioral impact

None intended.

The same Koin-managed dependencies are resolved when batch functionality needs them.

Only their construction timing changes.

---

# 4. Snapshot prerelease preferences once per repository reload

### Files

- `PatchBundleRepository.kt`
- `PatchBundleSource.kt`

### Problem

Bundle/source construction could repeatedly query the same prerelease-enabled source preference while processing multiple sources.

That introduces repeated preference access into a path that can execute many times during repository initialization/reload.

### Change

The prerelease-enabled source state is obtained once and reused throughout the source-construction operation.

Instead of conceptually doing:

    source A → read preference
    source B → read preference
    source C → read preference
    ...

the reload can do:

    read preference once
        ↓
    source A
    source B
    source C
    ...

### Benefit

This removes redundant blocking preference work from a repeated source-processing path.

The exact standalone wall-clock saving was not isolated independently, so it is included in the aggregate repository/reload improvement rather than assigned an artificial millisecond value.

---

# 5. Reuse already-loaded bundle manifest information

### Files

- `PatchBundleRepository.kt`
- `PatchBundleSource.kt`

### Problem

Repository processing already loads installed bundle metadata/manifests.

A later source-name synchronization path could perform another manifest lookup even though the relevant information had already been obtained earlier in the same processing pipeline.

### Change

The loaded manifest/name information is exposed and reused.

### Before

    load bundle metadata
        ↓
    read manifest
        ↓
    continue processing
        ↓
    name synchronization
        ↓
    read/scan manifest again

### After

    load bundle metadata
        ↓
    read manifest
        ↓
    retain required name information
        ↓
    name synchronization reuses it

### Benefit

Eliminates redundant bundle inspection from repository reload.

Again, this was not isolated as a standalone cold-start benchmark, so its contribution is represented by the aggregate metadata/reload measurements.

---

# 6. Bounded parallel bundle metadata loading

### File

- `PatchBundleRepository.kt`

### Problem

Bundle metadata operations are largely independent across sources.

Processing all of them serially unnecessarily extends repository reload time.

At the same time, launching unlimited parallel work would risk:

- excessive IO contention
- excessive CPU contention
- memory pressure
- poor behavior on devices with fewer resources

### Change

Metadata work is processed concurrently with an explicit concurrency bound.

This allows independent metadata operations to overlap while preventing unbounded fan-out.

Per-source failure handling remains intact.

### Measured result

The broader metadata phase improved from approximately:

    Baseline metadata:  ~945 ms
    Optimized metadata: ~851 ms

Difference:

    ~94 ms faster
    ~9.9% reduction

This is one of the clearest measured improvements in the repository pipeline.

The ~94 ms should be considered the improvement of the optimized metadata/repository path as a whole, rather than attributed exclusively to parallelism, because other metadata-path optimizations are present in this PR as well.

---

# 7. Bounded concurrent remote-source processing

### File

- `PatchBundleRepository.kt`

### Problem

Remote source updates are independent enough to benefit from concurrency, but unrestricted concurrency can cause excessive simultaneous:

- network requests
- source processing
- metadata work
- repository mutations

### Change

Remote-source update processing now uses bounded concurrency.

This allows multiple independent updates to overlap while maintaining a controlled maximum amount of simultaneous work.

Existing behavior is preserved for:

- cancellation
- progress reporting
- failure handling
- source updates
- final repository reload
- update application

### Benefit

The optimized repository reload path measured:

    Baseline reload:  ~1091 ms
    Optimized reload: ~974 ms

Difference:

    ~117 ms faster
    ~10.7% reduction

As with metadata, this number represents the optimized reload pipeline as a whole and should not be interpreted as 117 ms coming exclusively from this single concurrency change.

---

# 8. Persistent patch validation/index reuse

### File

- `PatchBundle.kt`

### Problem

Patch bundles that have already been validated can incur repeated validation/index work even when the underlying installed patch JAR has not changed.

This repeats work whose result is already known.

### Change

The patch loader can reuse persistent validation/index information when the bundle file still matches the state associated with the cached result.

Conceptually:

Before:

    load bundle
        ↓
    validate bundle
        ↓
    inspect/index patches
        ↓
    load

Repeated again on subsequent loads.

After:

    load bundle
        ↓
    check persistent validation/index state
        ↓
    unchanged?
       ↙       ↘
     yes        no
      ↓          ↓
    reuse      normal validation/index
      ↓          ↓
          load

### Safety

Cache reuse is conditional on the underlying bundle state.

Changed bundles continue through normal validation/loading rather than blindly trusting stale cached information.

### Benefit

This reduces repeated patch-loader work for unchanged installed bundles.

A standalone cold-start delta was not isolated for this optimization, so it is included in the aggregate repository improvements.

---

# Overall measured impact

The most important end-to-end result was:

    Activity → Home first

    Baseline:  ~698 ms
    Optimized: ~548 ms

    Saved:     ~150 ms
    Reduction: ~21.5%

Repository pipeline measurements also improved:

    Metadata
    ~945 ms → ~851 ms
    ~94 ms saved (~9.9%)

    Reload
    ~1091 ms → ~974 ms
    ~117 ms saved (~10.7%)

Meanwhile:

    Activity → first frame
    ~49 ms → ~44 ms

remained approximately stable.

This indicates that the optimization primarily reduces/deconflicts application and repository work occurring around Home startup rather than substantially altering Android's initial frame delivery.

---

# Controlled negative experiments

Several plausible follow-up optimizations were tested and intentionally **not included** because they did not improve the controlled P40 keeper.

### Defer launcher shortcut observer

    Keeper:      548 ms
    Experiment:  582 ms
    Regression:  +34 ms

Rejected.

### Defer avatar PatchBundleRepository.sources access

    Keeper:      548 ms
    Experiment:  562 ms
    Regression:  +14 ms

Rejected.

### Defer BlocklistRepository startup

    Keeper:      548 ms
    Experiment:  586 ms
    Regression:  +38 ms

Rejected.

### Defer all three paths together

    Keeper:      548 ms
    Experiment:  597 ms
    Regression:  +49 ms

Rejected.

These experiments were useful because they demonstrated that simply moving additional repository consumers behind the first-frame boundary does not automatically improve startup.

None of those experimental behavior changes are present in this PR.

---

# Additional Compose profiling

After reaching the optimized keeper, additional profiling investigated the remaining Home composition cost.

The first actual visible LazyColumn item lambda measured only approximately:

    ~3.8 ms

and the LazyColumn call itself approximately:

    ~5.6 ms

These values were too small to justify risky UI restructuring.

No speculative Compose optimization from those investigations is included here.

---

# Validation

The final production branch was cleaned of all profiling instrumentation before submission.

Removed development-only instrumentation includes:

- `MorpheStartupP*` markers
- `MorphePerf`
- `MorphePerfMeta`
- profiling-only `System.nanoTime()` measurements
- startup timing log statements
- instrumentation-only counters
- experimental profiling code

Verification:

    git grep -n -E \
    "MorpheStartupP(21|24|25|26|27|28|40)|MorphePerf|MorphePerfMeta"

returned no matches.

Final production diff:

    8 files changed
    239 insertions
    74 deletions

Production files:

- ManagerApplication.kt
- ViewModelModule.kt
- PatchBundleSource.kt
- PatchBundleRepository.kt
- PatchBundle.kt
- HomeScreen.kt
- HomeViewModel.kt
- MainViewModel.kt

Build validation:

    ./gradlew.bat :app:assembleDebug

Result:

    BUILD SUCCESSFUL

The final submitted branch contains the production optimization behavior only; profiling and unsuccessful experimental changes are not included.

### DIFF
--- a/app/src/main/java/app/morphe/manager/di/ViewModelModule.kt
+++ b/app/src/main/java/app/morphe/manager/di/ViewModelModule.kt
@@ -1,12 +1,36 @@
 package app.morphe.manager.di
 
+import app.morphe.manager.domain.repository.PatchBundleRepository
 import app.morphe.manager.ui.viewmodel.*
 import org.koin.core.module.dsl.viewModelOf
 import org.koin.dsl.module
+import org.koin.androidx.viewmodel.dsl.viewModel
 
 val viewModelModule = module {
     viewModelOf(::MainViewModel)
-    viewModelOf(::HomeViewModel)
+    viewModel {
+        HomeViewModel(
+            get(),
+            lazy(LazyThreadSafetyMode.SYNCHRONIZED) { get<PatchBundleRepository>() },
+            get(),
+            get(),
+            get(),
+            get(),
+            get(),
+            get(),
+            get(),
+            get(),
+            get(),
+            get(),
+            get(),
+            get(),
+            get(),
+            get(),
+            get(),
+            get(),
+            get()
+        )
+    }
     viewModelOf(::ThemeSettingsViewModel)
     viewModelOf(::SettingsViewModel)
     viewModelOf(::PatcherViewModel)
--- a/app/src/main/java/app/morphe/manager/ManagerApplication.kt
+++ b/app/src/main/java/app/morphe/manager/ManagerApplication.kt
@@ -29,6 +29,7 @@ import coil.ImageLoader
 import com.google.android.gms.common.ConnectionResult
 import com.google.android.gms.common.GoogleApiAvailability
 import com.topjohnwu.superuser.Shell
+import kotlinx.coroutines.CompletableDeferred
 import kotlinx.coroutines.Dispatchers
 import kotlinx.coroutines.MainScope
 import kotlinx.coroutines.flow.first
@@ -66,6 +67,19 @@ class ManagerApplication : Application() {
         private const val SHORTCUT_ICON_PX = 192
     }
     private val scope = MainScope()
+
+    // P40: nothing may construct PatchBundleRepository until Home has rendered one frame.
+    private val homeFirstFrameReady = CompletableDeferred<Unit>()
+    private val patchBundleRepositoryConstructed = CompletableDeferred<Unit>()
+
+    fun signalHomeFirstFrameReady() {
+        if (homeFirstFrameReady.complete(Unit)) {
+        }
+    }
+
+    suspend fun awaitPatchBundleRepositoryConstructed() {
+        patchBundleRepositoryConstructed.await()
+    }
     private val prefs: PreferencesManager by inject()
     private val patchBundleRepository: PatchBundleRepository by inject()
     private val blocklistRepository: BlocklistRepository by inject()
@@ -164,8 +178,15 @@ class ManagerApplication : Application() {
 
         // First touch of the repository builds the Ktor client, which costs seconds on a cold
         // start, so it happens here on a background dispatcher rather than in the Koin graph
+        // P40: preserve the same startup work, but do not compete with Activity/Compose startup.
+        // Home signals after its first rendered frame; construction then happens on Default.
         scope.launch(Dispatchers.Default) {
-            with(patchBundleRepository) {
+            homeFirstFrameReady.await()
+
+            val repository = patchBundleRepository
+            patchBundleRepositoryConstructed.complete(Unit)
+
+            with(repository) {
                 reload()
                 updateCheck()
             }
@@ -177,12 +198,14 @@ class ManagerApplication : Application() {
         scope.launch(Dispatchers.Default) {
             blocklistRepository.loadFromCache()
             blocklistRepository.refresh()
+            patchBundleRepositoryConstructed.await()
             patchBundleRepository.logBlockedSources()
         }
 
         // Preload bundle avatar images into AvatarCache while the user hasn't opened the sheet yet.
         // Suspends until sources are ready, then fetches all URLs in parallel on IO threads
         scope.launch(Dispatchers.IO) {
+            patchBundleRepositoryConstructed.await()
             patchBundleRepository.sources.first { it.isNotEmpty() }.forEach { bundle ->
                 launch {
                     val avatarUrls = bundle.avatarUrls
--- a/app/src/main/java/app/morphe/manager/domain/bundles/PatchBundleSource.kt
+++ b/app/src/main/java/app/morphe/manager/domain/bundles/PatchBundleSource.kt
@@ -48,7 +48,8 @@ sealed class PatchBundleSource(
     }
 
     val version get() = manifestAttributes?.version
-    val isNameOutOfDate get() = manifestAttributes?.name?.let { it != name } == true
+    internal val installedManifestName get() = manifestAttributes?.name
+    val isNameOutOfDate get() = installedManifestName?.let { it != name } == true
     val error get() = (state as? State.Failed)?.throwable
     val displayTitle get() = displayName?.takeUnless { it.isBlank() } ?: name
 
--- a/app/src/main/java/app/morphe/manager/domain/repository/PatchBundleRepository.kt
+++ b/app/src/main/java/app/morphe/manager/domain/repository/PatchBundleRepository.kt
@@ -34,7 +34,9 @@ import kotlinx.collections.immutable.toPersistentMap
 import kotlinx.coroutines.*
 import kotlinx.coroutines.flow.*
 import kotlinx.coroutines.sync.Mutex
+import kotlinx.coroutines.sync.Semaphore
 import kotlinx.coroutines.sync.withLock
+import kotlinx.coroutines.sync.withPermit
 import java.io.File
 import java.io.FileInputStream
 import java.io.InputStream
@@ -62,6 +64,15 @@ class PatchBundleRepository(
 
     private val scope = CoroutineScope(Dispatchers.Default)
     private val store = Store<BundleState>(scope, BundleState.Loading)
+
+    // P1 instrumentation only. No scheduling, concurrency, or networking behavior is changed.
+
+    private fun recordPeak(peak: AtomicInteger, value: Int) {
+        while (true) {
+            val current = peak.get()
+            if (value <= current || peak.compareAndSet(current, value)) return
+        }
+    }
 
     val bundleState: StateFlow<BundleState> = store.state
         .stateIn(scope, SharingStarted.Eagerly, BundleState.Loading)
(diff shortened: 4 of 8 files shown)


app/src/main/java/app/morphe/manager/di/ViewModelModule.kt
```diff
@@ -1,12 +1,36 @@
...
+            get(),
+            get(),
+            get(),
+            get(),
+            get(),
+            get(),
+            get(),
+            get()
+        )
+    }
```
LisoUseInAIKyrios, line 33: This adds code complexity and a little bit of "Wut? Huh?"

100ms is not significant for app startup. Manager is not something the user typically opens multiple times in a day (at least not opening at the same frequency as a browser, email app, or other more common apps).

jo2dan94:
ok you dont have to be rude about it. but with all these optimizations i did find my 141 sources loading alot faster. i didnt say u had to accept the pull request 🙄

Sent from [Proton Mail](https://proton.me/mail/home) for Android.

-------- Original Message --------

LisoUseInAIKyrios:
Not being rude, just being honest.

LisoUseInAIKyrios:
Even if this was a simple change, it changes the app loading to be non-deterministic due to the concurrent nature.

A future concurrency bug could be introduced by accident (or there could be a bug introduced now because of this), and the bug could only manifest sometimes or for certain devices.

I don't think it's worth saving 100ms because of this.

MarcaDian:
Thanks for the detailed writeup and the profiling effort. I'm closing this PR, but the notes below should be useful if you want to bring back the parts that hold up.

**1. The first-frame gate is bypassed before the first frame.**
`MainActivity.kt:528` reads `homeViewModel.bundleUpdateProgress` inside `composable<HomeScreen> { }` during the first composition, which forces the `lazy`. `HomeViewModel.init` also starts `observeLoadingState()`, `observeInstalledAppUpdates()` and `observeSnackbarState()` in `viewModelScope` (Main.immediate), and all three touch `patchBundleRepository` immediately. So the repository is still constructed before the first frame, only now it happens on the main thread, whereas before it was built on `Dispatchers.Default` from `ManagerApplication`. That's a regression rather than a deferral, and it explains why "first frame" barely moved while the "Home first" marker did.

**2. Optimization №8 (validation/index reuse) is dead code.**
`.morphe-dex-class-index-v1` is never written anywhere. `git grep` finds only the two reads you added, so `canReuseValidation` is always `false`, and `classIndexFile` / `classIndexSignature` / `classIndexHit` are computed and never used. Net effect is two extra file reads per bundle load.

**3. Optimization №7 (bounded remote concurrency) is a no-op.**
`Semaphore(32)` wraps work already running on `Dispatchers.IO.limitedParallelism(4)`, so the permit never blocks. The `withPermit` body also wasn't re-indented.

**4. Instrumentation is still in the branch.**
The Validation section greps for `MorpheStartupP(21|24|…)|MorphePerf`, which is why it came back clean, but these remain:
- `PatchBundleRepository.kt:70` `recordPeak()`, unused, commented "P1 instrumentation only"
- `HomeScreen.kt:65` empty `LaunchedEffect(Unit) { }`
- `HomeScreen.kt:123-128` `p21ReadyLogged` plus a `LaunchedEffect` that only flips a flag
- `ManagerApplication.kt:75` `if (homeFirstFrameReady.complete(Unit)) { }` with an empty body
- `PatchBundleRepository.kt` empty `} else { }`, plus `val connected = …; if (!connected)` and `val ready = …; return ready`
- `PatchBundle.kt` `metadataFor()` expanded into three statements for no reason
- `p40RepositoryReady`, `p21ReadyLogged`, and the `P40` / `P21` / `P9` / `P4` / `P1` comments

**5. DI regressions.**
Replacing `viewModelOf(::HomeViewModel)` with 18 positional `get()` calls trades compile-time safety for one lazy parameter, so any constructor reorder now breaks silently at runtime. And `MainViewModel` reaching for `GlobalContext.get().get<…>()` is a service locator that bypasses Koin scoping and breaks testability. Pass `Lazy<T>` through the module instead, the same way you did for `HomeViewModel`.

**6. Undocumented behavior change in name sync.**
The old path used `src.patchBundle?.manifestAttributes?.name`, which skipped failed bundles. `installedManifestName` falls back to reading `patches.jar` even when `patchBundle` is null, so failed bundles now get their names synced too. The description says the opposite. Dropping the nested `dispatchAction("Sync names")` inside `doReload` is probably an improvement, but it should be described as the semantic change it is.

**7. Parallel metadata loading needs a lower bound.**
Eight concurrent `PatchBundle.Loader.metadata()` calls means eight DEX class-loading operations in flight, which is a real OOM risk on low-end devices. Also `limitedParallelism(8)` plus `Semaphore(8)` double-bounds the same thing, and the dispatcher view is recreated on every `loadMetadata` call instead of being a shared `val`.

**8. Diff noise.** `HomeScreen.kt` drops about six blank lines between logical blocks, adds a redundant `androidx.compose.runtime.remember` import alongside the existing `runtime.*`, and puts the `ManagerApplication` import above the `android.*` block.

**Worth keeping:** №4 (snapshotting `bundlePrereleasesEnabled` once) is a genuine win, just drop the now-unused default value on `load()`. #6 is fine with a bound of 2 to 4 and a single shared dispatcher. №5 works if you restore the `patchBundle != null` condition.

If you want to open a follow-up, I'd suggest a small PR with only those three, no instrumentation, no DI changes. For startup itself, removing the eager access at `MainActivity.kt:528` and in `HomeViewModel.init` while leaving construction in `ManagerApplication` on `Dispatchers.Default` would likely get most of the win with far less machinery.

PR #892, opened 2026-08-23
chore: Fix incorrect use of possessive "'s"

SehrGuterCode:
This annoys me every time I'm applying patches, so I finally decided to open a PR to fix it :)

### DIFF
--- a/app/src/main/res/values/strings.xml
+++ b/app/src/main/res/values/strings.xml
@@ -537,7 +537,7 @@ Uninstalling will permanently erase all app data"</string>
     <string name="patcher_message_4">Brewing digital magic… ☕</string>
     <string name="patcher_message_5">Coffee break recommended… ☕</string>
     <string name="patcher_message_6">Look outside, it\'s nice out there… 🌤️</string>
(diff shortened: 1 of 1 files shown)


LisoUseInAIKyrios:
It's not possessive, it's describing plural numbers.

https://www.grammarbook.com/blog/apostrophes/the-apostrophe-with-numbers-letters-and-abbreviations/

Thanks for watching out

SehrGuterCode:
@LisoUseInAIKyrios thanks for the link, but it seems to state the opposite of what you're saying:

> To form the plural of a single-digit number, add an s: Her phone number has four 7s in it.

> If you wish to form the possessive of a single number, add an apostrophe and an s: The number 7’s influence on my luck this month has been uncanny.

So based on my interpretation of this text, `1s and 0s` is still correct, as we don't want a possessive here, but instead plural numbers, which are formed without apostrophes.

LisoUseInAIKyrios:
Yes this change is correct. I only glanced at the change and mixed up what direction the change was from.

SehrGuterCode:
Thank you :)

PR #904, opened 2026-08-26
fix: Add "Pre-release updates" confirmation dialog

kitadai31:
The current description makes it difficult to understand that the "Pre-release updates" setting applies only to the Manager.
It can be misinterpret as "This option enables pre-release for each source in patch sources"
In fact, even the person who translated this string at Crowdin was confused and translated it as if this setting also enabled pre-releases of patches in my language.
This PR reduces confusion for users when enabling patches pre-releases.

Additionally, to make it even clearer, in addition to this change, I think we can also change the setting title to "Pre-release updates for Manager".

### DIFF
--- a/app/src/main/java/app/morphe/manager/ui/screen/settings/advanced/UpdatesSection.kt
+++ b/app/src/main/java/app/morphe/manager/ui/screen/settings/advanced/UpdatesSection.kt
@@ -47,6 +47,16 @@ fun UpdatesSettingsItem(
     val useManagerPrereleases by prefs.useManagerPrereleases.getAsState()
     val usePatchesPrereleases by prefs.bundlePrereleasesEnabled.getAsState()
     val showIntervalDialog = remember { mutableStateOf(false) }
+    val showPrereleaseWarning = remember { mutableStateOf(false) }
+
+    fun applyManagerPrereleases() {
+        settingsViewModel.toggleManagerPrereleases(
+            currentValue = useManagerPrereleases,
+            backgroundNotificationsEnabled = backgroundUpdateNotifications,
+            patchesPrereleaseIds = usePatchesPrereleases,
+            onCheckUpdate = onManagerPrereleasesToggle
+        )
+    }
 
     if (showIntervalDialog.value) {
         UpdateCheckIntervalDialog(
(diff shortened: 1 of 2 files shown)


MarcaDian:
We wanted to avoid the word "manager" when referring to the quasi-ecosystem, but I agree, it confuses users.

It’s also possible your translator simply didn't update the translation. Originally, there was a single toggle for both the manager and the patches; however, once we added support for other bundles, we needed a toggle for each one, so it was moved to the sources bottom sheet.

LisoUseInAIKyrios:
If it drops "Manager" it should be an improvement:
> Early access to new Morphe versions. Note: To enable pre-release patches, enable them in the patch sources list

The prior project used "Manager" in the app name and it caused a lot of confusion for new users. Users didn't understand why they needed to install "Manager".

Users looking to download "Morphe" should download and install an app named "Morphe", and they should tell their friends about "Morphe".

The name 'Manager' is used only here on GitHub to identify this subproject.

kitadai31:
> It’s also possible your translator simply didn't update the translation.

Upon checking, I found that it was an AI translation rather than a human one. It is possible that the AI carried over memory from the previous string.
The mystery is solved, thank you.

LisoUseInAIKyrios:
We had to change to a cheaper distilled AI model because Crowdin kept misbehaving and running excessive translation requests for no clear reason.

This month Crowdin AI ran up a bill higher than the 7 prior months combined. Changing to a cheaper model (that also has higher provider availability) has brought the cost back under control.

kitadai31:
😢

But please don't worry. That AI translation was done three months ago, so it's not due to the AI model being downgraded.

MarcaDian:
If necessary, I can add a confirmation dialog to this toggle, providing more details, for removing the note from the summary.

MarcaDian:
Since the latest changes to the manager add a confirmation dialog to the pre-release sources toggle, I added a similar one for the manager that explains exactly what it does. Please check if everything looks right, as I am not a native speaker. I also changed the string keys and added notes for translators to clarify the translation.

kitadai31:
LGTM.
I think this is a good change that improves consistency with the recently added confirmation dialog.
Thank you!

PR #906, opened 2026-08-27
perf: Repackage the archive only when signing rejects it

rotundtapir:
## Summary

Reduces on-device patch time, with no change to the patched output.

### Repackage the archive only when signing rejects it (`KeystoreManager`)

`sanitizeZipIfNeeded` exists for third-party APKs whose ZIP headers upset `ApkSigner`, but it ran
unconditionally, so every patch inflated and re-deflated the whole archive before signing. Anything
the patcher itself just wrote is already well-formed, making that work redundant in the common case.

This attempts the signature directly and falls back to repackaging only if the signer actually
throws. A malformed input is still handled; a well-formed one no longer pays for a full re-deflate
of the archive. How much time that saves depends on the archive size, so it varies by app.

## Correctness

The patched APK is content-identical to one produced by unmodified `dev`: same entry count, every
entry byte-identical in content, differing only in the generated `PatchInfo` timestamp and the
signature. The patched app installs, launches, and runs patched code.

The change does not alter patch semantics, compression, or the memory limit.

## Measurement

One illustrative run; these are not a promise, and actual savings depend heavily on the device, the
app being patched, and the archive size. Patching YouTube 21.04.223 on an emulator, patcher limit
1024 MB, the signing phase went from ~20s to ~4s once repackaging became conditional, because most
of that phase had been the redundant re-deflate rather than signing itself.

## Notes

These changes were developed and measured with AI assistance.

### DIFF
--- a/app/src/main/java/app/morphe/manager/domain/manager/KeystoreManager.kt
+++ b/app/src/main/java/app/morphe/manager/domain/manager/KeystoreManager.kt
@@ -45,16 +45,34 @@ class KeystoreManager(app: Application, private val prefs: PreferencesManager) {
         password = prefs.keystorePass.get()
     )
 
+    /**
+     * Signs [input] into [output].
+     *
+     * Repackaging the archive first fixes the malformed headers some third-party APKs carry, but it
+     * means inflating and re-deflating every entry of the archive, which is wasted whenever the
+     * archive was already well-formed - as it is for anything the patcher itself just wrote. So sign
+     * directly and fall back to [sanitizeZipIfNeeded] only if the signer actually rejects the input.
+     */
     suspend fun sign(input: File, output: File) = withContext(Dispatchers.Default) {
-        val sanitized = sanitizeZipIfNeeded(input)
-        ApkUtils.signApk(sanitized, output, prefs.keystoreAlias.get(), signingDetails())
-        if (sanitized != input) sanitized.delete()
+        val alias = prefs.keystoreAlias.get()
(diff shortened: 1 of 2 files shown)


MarcaDian:
Measured both changes on a Redmi (Android 15, 7.4 GB), patching YouTube 21.34.243 and Cake 6.4.0, a few runs per variant.

The signing change holds up. Signing YouTube is ~7s for a 110 MB output now, and the archive isn't deflated twice. I pushed a small follow-up on top: the fallback now only fires when the signer rejected the archive itself, so a bad keystore or a full disk doesn't pay for a full inflate + re-deflate before reporting an error a repackage was never going to fix.

The heap commit I couldn't measure any effect from. At the default 1280 MB limit the runs are indistinguishable (441s vs 441s, against a 31s spread between two runs of the same build). So I raised the pressure and set the limit to 768 MB, which is what a 3 GB phone gets. Allocation stalls do appear there, 22 of them, ~22s, about 5% of the run, but it's 21 with the commit and 22 without. They start around the six minute mark, once the heap is already against the ceiling, and starting it bigger only skips the early ramp.

Also worth knowing: on this device heapminfree, heapmaxfree and heaptargetutilization are already 8m, 32m and 0.5, so three of the four lines set what is already set.

Reverted it in a separate commit rather than touching yours. The problem you were going after is real though. 22s of stalls on a tight limit is worth chasing, it just doesn't look like the starting heap size is the lever.

PR #913, opened 2026-08-30
fit: Handle architecture discovery in malformed splits

hoo-dles:
`¯\_(ツ)_/¯`


MarcaDian:
The same gap was fixed in `dev` a few minutes before this PR was opened 😅 https://github.com/MorpheApp/morphe-manager/commit/f89129a00cc0f76f8cd3195e10e942147e410342

PR #940, opened 2026-09-04
Change, ProcessRuntime.kt: allow higher memory usage

### LINKED ISSUE #939: feat: allow to increase heap to way more than 1280MB
### Feature description

I frequently run into crashes trying to apply a few patches to games. My phone has enough storage so why not allow a way bigger memory heap than 1GB?

### Motivation

Prevent frequent crashes for many apps and multiple patches

### Acknowledgements

secretmango:
Fixes #939

Is a change in `PROCESS_RUNTIME_MEMORY_MAX_LIMIT_INITIALIZATION` also needed to prevent crashes?

### DIFF
--- a/app/src/main/java/app/morphe/manager/patcher/runtime/ProcessRuntime.kt
+++ b/app/src/main/java/app/morphe/manager/patcher/runtime/ProcessRuntime.kt
@@ -33,7 +33,7 @@ import kotlin.time.Duration.Companion.seconds
 // If too much memory is requested then some devices become extremely slow
 // for unknown reason (using flash memory as swap file?)
 const val PROCESS_RUNTIME_MEMORY_MINIMUM = 512
-const val PROCESS_RUNTIME_MEMORY_MAX_LIMIT = 1280
+const val PROCESS_RUNTIME_MEMORY_MAX_LIMIT = 2560
 const val PROCESS_RUNTIME_MEMORY_MAX_LIMIT_INITIALIZATION = 1024
 private const val PROCESS_RUNTIME_MEMORY_DEFAULT_MINIMUM = 640
 const val PROCESS_RUNTIME_MEMORY_LOW_WARNING = 640


secretmango:
Why tumbs down? The patching constantly hovered around 1200MB for me and consistently crashes with only a few patches applied. This would have the possibility to fix it wouldnt it? At least giving it a try, I would beta test on 2 phones

LisoUseInAIKyrios:
We previously allowed selecting up to 1600mb, but we reduced it to 1280mb for a reason.

As I explained in the issue you created only a few minutes before you opening this PR, some devices become dramatically slower when high amounts of memory are requested. This issue still exists even with 1280mb and some weird devices require manually changing to 768mb or 512mb of memory or patching takes forever or fails with memory errors.

Post a Manager expert mode screenshot showing 1280mb of memory being maxed out.

secretmango:
signal-2026-09-04-20-24-11-920

Here for Jetpack joyride, same happens in the Subway Surfers scenario I described in the issue. As soon as it hits 1280 it crashes. The huge memory increase is generally at the stage where it processes all the classes in parallel. So might not even be related to the patches.

If some devices require manual configuration below 1280, that would be another argument why the max limit would not be that important.

These 2 pretty old games, both below 500MB each should be patchable. The low limit simply means that they are not patchable, which is an issue.

LisoUseInAIKyrios:
@secretmango Please list the exact patches you included.

It's possible one of the patches is using `mutableClassBy` when it doesn't need the mutable version.

(and this is one reason why opening a PR before any discussion stinks. The conversation is scattered across the issue and PR that were created minutes apart from each other).

secretmango:
sorry, lets continue in the issue

MarcaDian:
Superseded by bd98d0585d89646d8bc3dd33b1de00a89f5ed45f

PR #944, opened 2026-09-05
feat: fall back to the bundle website for unresolved downloads

catsmoker:
## Problem

The API answers a download search with a redirect chain, and apps it knows nothing about (for example Anime Witcher) dead-end on a plain Google search page. The download instructions dialog then tells the user to search the web instead of pointing at anywhere useful.

Patch bundles can already declare their own official site via the \Website\ manifest attribute, but nothing reads it.

## Change

When the redirect chain cannot land on a download page, fall back to the bundle's officially declared website before giving up to a web search.

- \PatchBundleSource\: expose \website\ from the bundle manifest attributes
- \PatchBundleRepository.websiteFor(uids)\: site of the first covering bundle that declares one, in preference order
- \DownloadUrlResolver.resolve(packageName, version, officialUrl)\: prefer the official site in every dead-end path (timeout, no redirect location, gave up on the redirect loop, a plain \google.com\ search page, and a Uptodown link that lost its build)
- \HomeViewModel\ / \BatchPatcherViewModel\: look up the covering bundle's website per pending app and pass it through, also using it as the download-helper \allbackWebUrl\

For apps whose bundle declares no website the behavior is unchanged (the site-restricted web search is still the fallback).

Verified with \:app:compileDebugKotlin\.

### DIFF
--- a/app/src/main/java/app/morphe/manager/domain/bundles/PatchBundleSource.kt
+++ b/app/src/main/java/app/morphe/manager/domain/bundles/PatchBundleSource.kt
@@ -48,6 +48,10 @@ sealed class PatchBundleSource(
     }
 
     val version get() = manifestAttributes?.version
+
+    /** Official site the bundle publishes its app from, declared in the bundle manifest. */
+    val website get() = manifestAttributes?.website
+
     val isNameOutOfDate get() = manifestAttributes?.name?.let { it != name } == true
     val error get() = (state as? State.Failed)?.throwable
     val displayTitle get() = displayName?.takeUnless { it.isBlank() } ?: name
--- a/app/src/main/java/app/morphe/manager/domain/manager/DownloadUrlResolver.kt
+++ b/app/src/main/java/app/morphe/manager/domain/manager/DownloadUrlResolver.kt
@@ -23,38 +23,42 @@ import kotlin.time.Duration.Companion.seconds
 class DownloadUrlResolver(private val morpheAPI: MorpheAPI) {
 
     /**
-     * Resolves the download page for [packageName] at [version], falling back to a plain web
-     * search when the API cannot point anywhere useful.
+     * Resolves the download page for [packageName] at [version], falling back to the bundle's own
+     * website when it declared one, or to a plain web search otherwise, whenever the API cannot
+     * point anywhere useful.
+     *
+     * @param officialUrl The app's official site from the covering patch bundle, or null when the
+     *   bundle does not declare one.
      *
      * Every redirect along the way is followed: the API sometimes points at itself once more,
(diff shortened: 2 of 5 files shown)


LisoUseInAIKyrios:
This is assuming the bundle website has information on where to download an app from, which very few do.

For 99% of use cases, google search is the correct place to go as it includes sites that we don't index due to users preference such as APKPure and apkcombo, but are sometimes the only place to get the original APK.

LisoUseInAIKyrios:
Not related to this, but your repo is now broken in Morphe Manager because you removed the required patch bundle JSON files from the project root folder.

LisoUseInAIKyrios:
Users are also apprehensive to downloading random apks from GitHub repos. Users prefer downloading from apkmirror or another established site that validates the apks are original.

Why not upload these hard to find Witcher apks to apkmirror?

catsmoker:
I agree that Google should stay as the fallback. That's because Google can find sites like APKMirror, APKPure and APKCombo—places users might already. Prefer.

This change isn't meant to replace that behavior. It’s for a specific situation. That happens when a bundle already knows the source of an app but the API redirect ends up on a useless generic search page.

Take something like Anime Witcher. The bundle can already list its website. In that case sending the user to the known official site makes more sense than sending them to a Google search and asking them to search again.

I also agree that just listing a website doesn’t mean the APK can be downloaded from there. A better long-term fix might be to let a patch bundle specify a direct download page or even a URL template.. We keep Google as the fallback if neither the website nor a download link is available.

So I see the bundle’s Website field as a targeted fallback—not a replacement, for Google.

Yes uploading the Witcher APKs to APKMirror would be simple.. That means someone must manually upload and update them every time a new version comes out.

There’s another problem. Morphe’s current search system relies on package name, version, ABI and known APK sites. For example:

com.anime.witcher 1.4.8 arm64-v8a apkmirror.com OR uptodown.com OR apkpure.com OR apkcombo.com

Even if the app’s official site has the APK that page isn’t part of the search targets. So it might never show up.

That’s why the bundle’s Website field still matters. It doesn’t need to take over Google or the existing APK-site search. It just gives Morphe a place to send the user when the resolver already knows the official source—but the limited search didn’t find a downloadable copy.

For Witcher this avoids having to create and maintain an APK mirror just to make the patching process work.

MarcaDian:
A proper implementation requires a separate (optional) field in the bundle manifest.

I checked, and almost all existing bundles have flawed, clumsy links, some pointing to GitHub, others using deep links etc...

The current PR will break functionality for almost everyone who runs into issues with the APK's availability on APKMirror or another site.

MarcaDian:
We’ll come back to this later, if necessary.

Thanks for the time.

